PR #18, opened 2023-03-10
Feature: Uses could not use custom timeout

hoh:


src/aleph/sdk/client.py
```diff
@@ -443,9 +443,11 @@ class AlephClient:
     api_server: str
     http_session: aiohttp.ClientSession
 
-    def __init__(self, api_server: str):
+    def __init__(
+        self, api_server: str, timeout: Optional[aiohttp.ClientTimeout] = None
+    ):
         self.api_server = api_server
-        self.http_session = aiohttp.ClientSession(base_url=api_server)
+        self.http_session = aiohttp.ClientSession(base_url=api_server, timeout=timeout)
```
odesenfans, line 450: You can't pass `None` as a default, there's a sentinel object in the code of aiohttp. It's gonna have to be something like this:

```suggestion
        params = {"base_url": api_server}
        if timeout:
            params["timeout"] = timeout
        self.http_server = aiohttp.ClientSession(**params)
```

odesenfans (changes requested):

src/aleph/sdk/client.py
hoh: What about `timeout = timeout or aiohttp.client.sentinel` ?

hoh:
Replaced in  #24

PR #19, opened 2023-04-01
Feature: Signature verification for Solana and Ethereum messages

MHHukiewitz:
Problem: Users cannot verify signatures

Solution: Implement verify_signature() functions for **Ethereum** and **Solana** signatures, more are following.

Partly solves: #9

PR #20, opened 2023-04-07
Cosmos signature verification and fix

MHHukiewitz:
Cosmos code hasn't been updated since 2020. I added a test file for CSDK to check whether the signature verification function works correctly and found that the signing did not work at all.

- I fixed it by making `CSDKAccount.public_key()` directly return a base64-encoded string, which is what is being stored inside the signature dictionary.

- Added the `verifiy_signature()` function as in #19.

- Added `ecdsa` and `cosmospy` to the testing requirements.


MHHukiewitz:
This branch is based on #19; therefore it should be merged after #19.

MHHukiewitz:
Regarding encoding bytes with base64: https://github.com/cosmos/cosmos-sdk/pull/12910#discussion_r943984849

Cosmos did, or will do, a transition from base64 to hex bytes, similar to Ethereum

MHHukiewitz:
Closing for now, as it is not needed and maintaining our own branch of the cosmos-sdk might not seem worth it.

PR #22, opened 2023-04-19
Fix: Relay startup event through AlephApp

MHHukiewitz:
Problem: @app.on_event("startup") would not be triggered on uvicorn startup

Solution: Relay "lifespan" events from the AlephApp to the http_app

### DIFF
--- a/src/aleph/sdk/vm/app.py
+++ b/src/aleph/sdk/vm/app.py
@@ -62,7 +62,7 @@ class AlephApp:
         receive: Optional[Callable[[], Awaitable[Any]]] = None,
         send: Optional[Callable[[Dict[Any, Any]], Awaitable[Any]]] = None,
     ):
-        if scope["type"] in ("http", "websocket"):
+        if scope["type"] in ("http", "websocket", "lifespan"):
             if self.http_app:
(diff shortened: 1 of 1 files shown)


odesenfans:
I'm not sure where you find the `on_event` decorator on the `AlephApp` class? The fix looks correct but I don't 100% get your use case.

MHHukiewitz:
Works for me on this app: https://github.com/yourtrading-ai/fishnet-cod/blob/main/src/fishnet_cod/api/main.py#L93

MHHukiewitz:
Usecase is performing actions before the API gets ready. In this case its recreating the indices from POST messages with AARS

MHHukiewitz:
More on this topic is found here: https://fastapi.tiangolo.com/advanced/events/

odesenfans:
> Usecase is performing actions before the API gets ready. In this case its recreating the indices from POST messages with AARS

I meant: I don't see where the `on_event` method/decorator is implemented. It's a really useful feature indeed.

MHHukiewitz:
Here: https://fastapi.tiangolo.com/advanced/events/#startup-event
It somehow works

MHHukiewitz:
Can it be merged? @hoh

PR #24, opened 2023-05-12
Fix: AlephClient class could not use unix sockets

hoh:


src/aleph/sdk/client.py
```diff
@@ -443,9 +444,35 @@ class AlephClient:
...
+
+        connector: Optional[aiohttp.UnixConnector]
+        if api_unix_socket:
+            connector = aiohttp.UnixConnector(path=api_unix_socket)
+        elif api_unix_socket is None and settings.API_UNIX_SOCKET:
+            connector = aiohttp.UnixConnector(path=settings.API_UNIX_SOCKET)
+        else:
+            connector = None
+
+        if timeout:
```
odesenfans, line 468: I prefer a good old `kwargs = {"timeout": timeout}` in this case to reduce duplication.

src/aleph/sdk/client.py
```diff
@@ -443,9 +444,35 @@ class AlephClient:
...
     http_session: aiohttp.ClientSession
 
-    def __init__(self, api_server: str):
-        self.api_server = api_server
-        self.http_session = aiohttp.ClientSession(base_url=api_server)
+    def __init__(
+        self,
+        api_server: Optional[str],
+        timeout: Optional[aiohttp.ClientTimeout] = None,
+        api_unix_socket: Union[None, str, Literal[False]] = None,
```
odesenfans, line 451: This typing is a bit weird. In such cases I prefer to add a second option, like `allow_unix_socket: bool`.

src/aleph/sdk/client.py
```diff
@@ -443,9 +444,35 @@ class AlephClient:
...
+        elif settings.API_HOST:
+            self.api_server = settings.API_HOST
+        else:
+            raise ValueError("Missing API host")
+
+        connector: Optional[aiohttp.UnixConnector]
+        if api_unix_socket:
+            connector = aiohttp.UnixConnector(path=api_unix_socket)
+        elif api_unix_socket is None and settings.API_UNIX_SOCKET:
+            connector = aiohttp.UnixConnector(path=settings.API_UNIX_SOCKET)
```
odesenfans, line 464: Nitpick: little bit of duplication here, maybe extract the socket path before spawning the `aiohttp.UnixConnector` instance.

src/aleph/sdk/client.py
```diff
@@ -443,9 +444,35 @@ class AlephClient:
...
-    def __init__(self, api_server: str):
-        self.api_server = api_server
-        self.http_session = aiohttp.ClientSession(base_url=api_server)
+    def __init__(
+        self,
+        api_server: Optional[str],
+        timeout: Optional[aiohttp.ClientTimeout] = None,
+        api_unix_socket: Union[None, str, Literal[False]] = None,
+    ):
+        if api_server:
```
odesenfans, line 453: Could you add checks to verify that the options specified by the user make sense? Ex: they didn't specify `api_server` and `api_unix_socket` at the same time.

odesenfans (changes requested):

src/aleph/sdk/client.py
hoh: api_server is required by the HTTP protocol, even when using Unix sockets.

src/aleph/sdk/client.py
hoh: I propose another approach using `functools.partial`.

PR #25, opened 2023-05-12
Fix: No error was raised on rejected message

hoh:


src/aleph/sdk/client.py
```diff
@@ -1364,4 +1365,6 @@ async def submit(
             storage_engine=storage_engine,
         )
         message_status = await self._broadcast(message=message, sync=sync)
+        if message_status == MessageStatus.REJECTED:
+            raise RejectedMessageError("This message was rejected")
```
odesenfans, line 1369: Pretty sure this case is already handled here: https://github.com/aleph-im/aleph-sdk-python/blob/bedffbb9ef54311b980ac9fd7c918454b51aeb9d/src/aleph/sdk/client.py#L901

Maybe there's an issue with that code, but this PR duplicates the logic. Could you check?

odesenfans (changes requested):

src/aleph/sdk/client.py
hoh: Indeed.

PR #29, opened 2023-05-21
Fix: ETH verify_signature fails on checksummed addresses

MHHukiewitz:
Problem: As Ethereum addresses may contain checksum information in capitalizing their letters, comparing the input address and the resulting public_key of the signature verification process resulted in false negatives.

Solution: Casefold both the address and public key before comparing them. Added further tests to both ETH and SOL accounts to increase coverage and test for other unforeseen errors.


hoh (approved):
Thanks.

PR #30, opened 2023-05-22
Problem: Signatures generated with Python 3.11 are invalid

hoh:
The signature buffer is affected by the new formatting of string
enums.

Solution: use the value of the message type enum.

Co-authored-by: Olivier Desenfans <desenfans.olivier@gmail.com>

### DIFF
--- a/src/aleph/sdk/utils.py
+++ b/src/aleph/sdk/utils.py
@@ -1,9 +1,10 @@
 import errno
 import logging
 import os
+from enum import Enum
 from pathlib import Path
 from shutil import make_archive
-from typing import Tuple, Type
+from typing import Tuple, Type, Union
 from zipfile import BadZipFile, ZipFile
 
 from aleph_message.models import MessageType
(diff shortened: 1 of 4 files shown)


src/aleph/sdk/utils.py
```diff
@@ -76,3 +77,16 @@ def check_unix_socket_valid(unix_socket_path: str) -> bool:
...
+    """Returns the value of an Enum, or the string itself when passign a string.
+
+    Python 3.11 adds a new formatting of string enums.
+    `str(MyEnum.value)` becomes `MyEnum.value` instead of `value`.
+    """
+    if isinstance(obj, str):
+        return obj
+    else:
+        assert isinstance(obj, Enum)
+        return obj.value
```
odesenfans, line 92: This implementation will not work because you check inheritance from `str` before `Enum`.

```
class C(str, Enum):
  V = "v"

c = C.V
isinstance(c, str) # True
```

```suggestion
def enum_as_str(obj: Union[str, Enum]) -> str:
    """Returns the value of an Enum, or the string itself when passign a string.

    Python 3.11 adds a new formatting of string enums.
    `str(MyEnum.value)` becomes `MyEnum.value` instead of `value`.
    """
    if not isinstance(obj, str):
        raise TypeError(f"Unsupported enum type: {type(obj)}")
    
    if isinstance(obj, Enum):
        return obj.value

    return obj
```

odesenfans (changes requested):
The current implementation of `enum_as_str` is incorrect. Also, maybe add tests for Python 3.11 in the CI?

odesenfans:
I think you need to rebase your PR.

src/aleph/sdk/utils.py
```diff
@@ -76,3 +77,18 @@ def check_unix_socket_valid(unix_socket_path: str) -> bool:
             unix_socket_path,
         )
     return True
+
+
+def enum_as_str(obj: Union[str, Enum]) -> str:
+    """Returns the value of an Enum, or the string itself when passign a string.
```
odesenfans, line 83:
```suggestion
    """Returns the value of an Enum, or the string itself when passing a string.
```

odesenfans (dismissed):

PR #31, opened 2023-05-23
Feature: SQLite-based Message Cache

MHHukiewitz:
## Problem
Some apps require access to the same messages over and over again. This puts unnecessary strain on the CCN API and increases the latency of apps using Aleph Messages.

## Solution
This pull request proposes a solution by introducing a message cache implemented with Peewee and an SQLite database in the Aleph SDK. The message cache effectively stores frequently accessed messages, eliminating the need for repetitive CCN API calls. With the message cache in place, apps utilizing the Aleph SDK will experience enhanced efficiency and responsiveness in accessing messages.


MHHukiewitz:
After introducing the AlephClientInterface, I get a mypy error, that I really don't get: https://github.com/aleph-im/aleph-sdk-python/actions/runs/5071082316/jobs/9107016057?pr=31

Even though I am exactly copying the signature of the `watch_messages` function:
```py
    async def watch_messages(
        self,
        message_type: Optional[MessageType] = None,
        content_types: Optional[Iterable[str]] = None,
        refs: Optional[Iterable[str]] = None,
        addresses: Optional[Iterable[str]] = None,
        tags: Optional[Iterable[str]] = None,
        hashes: Optional[Iterable[str]] = None,
        channels: Optional[Iterable[str]] = None,
        chains: Optional[Iterable[str]] = None,
        start_date: Optional[Union[datetime, float]] = None,
        end_date: Optional[Union[datetime, float]] = None,
    ) -> AsyncIterable[AlephMessage]:
```

mypy raises:

```
src/aleph/sdk/cache.py:370:5: error: Return type "AsyncIterable[Any]" of "watch_messages" incompatible with return type "Coroutine[Any, Any, AsyncIterable[Any]]" in supertype "AlephClientInterface"  [override]
src/aleph/sdk/client.py:708:5: error: Return type "AsyncIterable[Any]" of "watch_messages" incompatible with return type "Coroutine[Any, Any, AsyncIterable[Any]]" in supertype "AlephClientInterface"  [override]
```

MHHukiewitz:
Looking into [snapshot testing](https://pypi.org/project/pytest-snapshot/) to figure out how to mock API responses.

src/aleph/sdk/cache.py
```diff
@@ -0,0 +1,529 @@
...
+
+
+class JSONDictEncoder(json.JSONEncoder):
+    def default(self, obj):
+        if isinstance(obj, BaseModel):
+            return obj.dict()
+        return json.JSONEncoder.default(self, obj)
+
+
+class PydanticField(Field, Generic[T]):
```
hoh, line 60: This is an interesting approach.

src/aleph/sdk/cache.py
```diff
@@ -0,0 +1,529 @@
...
+    """
+    A field for storing pydantic model types as JSON in a database. Uses json for serialization.
+    """
+
+    field_type = "text"
```
hoh, line 65: SQLite has support for JSON fields and there is a JSONField in Peewee for JSON content. Why don't you use it ?

http://docs.peewee-orm.com/en/latest/peewee/sqlite_ext.html#JSONField

src/aleph/sdk/cache.py
```diff
@@ -0,0 +1,529 @@
...
+            return None
+        return value.json()
+
+    def python_value(self, value: Optional[str]) -> Optional[T]:
+        if value is None:
+            return None
+        return self.type.parse_raw(value)
+
+
+class JsonField(Field):
```
hoh, line 83: See previous comment about http://docs.peewee-orm.com/en/latest/peewee/sqlite_ext.html#JSONField

src/aleph/sdk/cache.py
```diff
@@ -0,0 +1,529 @@
...
+            return None
+        return json.dumps(value, cls=JSONDictEncoder)
+
+    def python_value(self, value: Optional[str]) -> Optional[dict]:
+        if value is None:
+            return None
+        return json.loads(value)
+
+
+class MessageModel(Model):
```
hoh, line 101: Do we have a process to validate the mapping between these fields and those from aleph-message when the models evolve ?

src/aleph/sdk/cache.py
```diff
@@ -0,0 +1,529 @@
...
+    A wrapper around an sqlite3 database for storing AlephMessage objects.
+    """
+
+    def __init__(self):
+        if db.is_closed():
+            db.connect()
+            db.create_tables([MessageModel])
+
+    def __del__(self):
+        db.close()
```
hoh, line 143: Would it make sense to use a context manager instead to handle this automatically ?

src/aleph/sdk/cache.py
```diff
@@ -0,0 +1,529 @@
...
+
+    def __getitem__(self, item_hash) -> Optional[AlephMessage]:
```
hoh, line 145: What type(s) is `item_hash` supposed to be ?

src/aleph/sdk/cache.py
```diff
@@ -0,0 +1,529 @@
...
+
+        if conditions:
+            query = query.where(*conditions)
+
+        async for item in query:
+            yield model_to_message(item)
+
+
+def message_to_model(message: AlephMessage) -> Dict:
+    return {
```
hoh, line 423: Why don't you use `message.dict()` https://docs.pydantic.dev/latest/usage/exporting_models/ ?

src/aleph/sdk/cache.py
```diff
@@ -0,0 +1,529 @@
...
+        MessageModel.delete().where(MessageModel.item_hash == item_hash).execute()
+
+    def __contains__(self, item_hash):
+        return MessageModel.select().where(MessageModel.item_hash == item_hash).exists()
+
+    def __len__(self):
+        return MessageModel.select().count()
+
+    def __iter__(self):
+        return iter(MessageModel.select())
```
hoh, line 165: Can you describe how this work in a comment ?

src/aleph/sdk/cache.py
```diff
@@ -0,0 +1,529 @@
...
+
+    def __repr__(self):
+        return f"<MessageCache: {db}>"
+
+    def __str__(self):
+        return repr(self)
+
+    def add_many(self, messages: Union[AlephMessage, List[AlephMessage]]):
+        if not isinstance(messages, list):
```
hoh, line 174: Why allow adding a single message outside of an iterable ?

src/aleph/sdk/cache.py
```diff
@@ -0,0 +1,529 @@
...
+    def __iter__(self):
+        return iter(MessageModel.select())
+
+    def __repr__(self):
+        return f"<MessageCache: {db}>"
+
+    def __str__(self):
+        return repr(self)
+
+    def add_many(self, messages: Union[AlephMessage, List[AlephMessage]]):
```
hoh, line 173: Why not allow `Iterable[AlephMessage]` for generators etc ?

src/aleph/sdk/cache.py
```diff
@@ -0,0 +1,529 @@
...
+        if not isinstance(messages, list):
+            messages = [messages]
+
+        data_source = (message_to_model(message) for message in messages)
+        with db.atomic():
+            for batch in chunked(data_source, 100):
```
hoh, line 179: Can you add comments describing why you chunk and why by that size ?

src/aleph/sdk/cache.py
```diff
@@ -0,0 +1,529 @@
...
+                MessageModel.insert_many(batch).on_conflict_replace().execute()
+
+    def get_many(
+        self, item_hashes: Union[Union[ItemHash, str], List[Union[ItemHash, str]]]
```
hoh, line 183: Why allow `str` and not just `ItemHash` ? And why not require an iterable ?

src/aleph/sdk/cache.py
```diff
@@ -0,0 +1,529 @@
...
+        page: int = 1,
+        types: Optional[Iterable[str]] = None,
+        refs: Optional[Iterable[str]] = None,
+        addresses: Optional[Iterable[str]] = None,
+        tags: Optional[Iterable[str]] = None,
+        hashes: Optional[Iterable[str]] = None,
+        channels: Optional[Iterable[str]] = None,
+        chains: Optional[Iterable[str]] = None,
+        start_date: Optional[Union[datetime, float]] = None,
+        end_date: Optional[Union[datetime, float]] = None,
```
hoh, line 246: Would it make sense to add a `Filter` model that contains all the fields used to filter the query ? This would remove duplication with other methods below.

src/aleph/sdk/cache.py
```diff
@@ -0,0 +1,529 @@
...
+    if len(channels) == 1:
+        return MessageModel.channel == channels[0]
+    return MessageModel.channel.in_(channels)
+
+
+def query_chains(chains):
+    chains = list(chains)
+    if len(chains) == 1:
+        return MessageModel.chain == chains[0]
+    return MessageModel.chain.in_(chains)
```
hoh, line 529: Why so much code duplication ?

src/aleph/sdk/cache.py
```diff
@@ -0,0 +1,529 @@
...
+    elif item.type == MessageType.aggregate.value:
+        return AggregateMessage.parse_obj(item_dict)
+    elif item.type == MessageType.store.value:
+        return StoreMessage.parse_obj(item_dict)
+    elif item.type == MessageType.forget.value:
+        return ForgetMessage.parse_obj(item_dict)
+    elif item.type == MessageType.program.value:
+        return ProgramMessage.parse_obj(item_dict)
+    else:
+        raise ValueError(f"Unknown message type {item.type}")
```
hoh, line 473: This can certainly be done in a smarter way, maybe using `aleph_message.models.parse_message` ?
https://github.com/aleph-im/aleph-message/blob/main/aleph_message/models/__init__.py#L359

src/aleph/sdk/client.py
```diff
@@ -500,14 +501,6 @@ async def fetch_aggregate(
...
         limit: int = 100,
     ) -> Dict[str, Dict]:
-        """
-        Fetch a value from the aggregate store by owner address and item key.
-
-        :param address: Address of the owner of the aggregate
-        :param key: Key of the aggregate
-        :param limit: Maximum number of items to fetch (Default: 100)
-        """
-
```
hoh, line 510: Why delete all the docstrings ?

src/aleph/sdk/conf.py
```diff
@@ -33,6 +33,11 @@ class Settings(BaseSettings):
 
     CODE_USES_SQUASHFS: bool = which("mksquashfs") is not None  # True if command exists
 
+    CACHE_DB_PATH: Path = Field(
```
hoh, line 36: We can be very explicit here.
```suggestion
    CACHE_DATABASE_PATH: Path = Field(
```

src/aleph/sdk/interface.py
```diff
@@ -0,0 +1,466 @@
...
+    MessageType,
+    PostMessage,
+)
+from aleph_message.models.program import Encoding
+from aleph_message.status import MessageStatus
+
+from aleph.sdk.types import GenericMessage, StorageEnum
+
+
+class AlephClientInterface(ABC):
```
hoh, line 32: The term Interface in Python refers to class that do not implement their methods, which is confusing here since some of them do have an implementation.

tests/unit/test_cache.py
```diff
@@ -0,0 +1,61 @@
...
+    # test listen until first message
+    coro = cache.listen_to(auth_session.watch_messages())
+    task = asyncio.create_task(coro)
+    before = len(cache)
+    # send message
+    await auth_session.create_aggregate("test", {"test": "test"})
+    await asyncio.sleep(2)  # wait for message to be received
+    task.cancel()
+    after = len(cache)
+    assert after > before
```
hoh, line 61: What is the coverage of the tests ?

hoh (commented):
I added a first batch of comments.

src/aleph/sdk/cache.py
MHHukiewitz: Is it bad? :sweat_smile:

src/aleph/sdk/cache.py
MHHukiewitz: Ah yes, I see. I was first trying to use it but it threw exceptions with unserializable classes. Then I went for this approach, but I did not notice that I could define my custom Encoder for this JSONField class.

src/aleph/sdk/cache.py
MHHukiewitz: Nope. I thought about whether it would make sense to include them in aleph-message, but it doesn't seem ideal either.

src/aleph/sdk/cache.py
MHHukiewitz: It is an iterator over all items in MessageModel. But I just noticed that it should actually invoke `model_to_message()` with every yielded item.

src/aleph/sdk/cache.py
MHHukiewitz: Type annotations will be added.

src/aleph/sdk/cache.py
MHHukiewitz: Yes, I think so. Otherwise, we might end up in a confusing situation where multiple `MessageCache`s exist and deleting one of them would cause all the others to stop functioning.

src/aleph/sdk/cache.py
MHHukiewitz: Usability? Readability? Idk :man_shrugging:

src/aleph/sdk/cache.py
MHHukiewitz: Again, I think usability and flexibility. If a user decides not to use pydantic validation, there might also be reasons for it, no?

src/aleph/sdk/cache.py
MHHukiewitz: Yes, I thought about this too, similar to what we do in the TS SDK. But it would have been a breaking change; that's why I avoided it.

src/aleph/sdk/cache.py
MHHukiewitz: Because I needed to accommodate the special cases for which one might like an index on columns like `tags` or `ref` on the table, but were not "first-class citizens" in `AlephMessage`.

src/aleph/sdk/cache.py
MHHukiewitz: Hm yes, I see how that could be solved.

src/aleph/sdk/client.py
MHHukiewitz: They have been moved to the interface

tests/unit/test_cache.py
MHHukiewitz: Cache.py: 60% 
Node.py: 0%

MHHukiewitz:
Thanks @hoh, this helps prioritizing on which points work on next for me.

src/aleph/sdk/cache.py
MHHukiewitz: This is problematic though, if `:memory:` is being used as DB path. Closing the connection seems to reset the DB and forget all tables.

hoh: The singleton pattern of Peewee can byte us there. What about storing it as an attribute on the instance instead of having a global `db` object ?

MHHukiewitz: I resolved this problem by having an instance counter for the `MessageCache` which will delete `db`, if it is the last instance being deleted. It is still a global though, still seems the best way to work with it and peewee.

MHHukiewitz:
@hoh 
I'd like to point out that this line:
```py
if isinstance(messages, AlephMessage):
```
is making tests fail in Python versions =< 3.9. The alternative is making comparisons with every single Aleph message type (`PostMessage`, `AggregateMessage`, ...) which I think might be a bit ridiculous.

As Python progressed already to 3.11 now, can it be expected of people wanting to use our newest SDK versions to use at least Python 3.10?

hoh:
> @hoh I'd like to point out that this line:
> 
> ```python
> if isinstance(messages, AlephMessage):
> ```
> 
> is making tests fail in Python versions =< 3.9. The alternative is making comparisons with every single Aleph message type (`PostMessage`, `AggregateMessage`, ...) which I think might be a bit ridiculous.

The solution seems to be using `isinstance(data, typing.get_args(Constant))`

https://stackoverflow.com/a/73874277

> 
> As Python progressed already to 3.11 now, can it be expected of people wanting to use our newest SDK versions to use at least Python 3.10?

Not as long as we support Debian 11, which ships with Python 3.9.
The SDK should work on as many environments as possible, and keep doing so for a while. The situation is different for PyAleph since it is only expected to be deployed in containers and already uses Python 3.11.

MHHukiewitz:
Closed for now, as it will be broken down into smaller PRs

PR #32, opened 2023-05-30
Allow multiple message types

MHHukiewitz:
Client-side deprecation of `message_type`, see: https://github.com/aleph-im/pyaleph/pull/444

### DIFF
--- a/src/aleph/sdk/client.py
+++ b/src/aleph/sdk/client.py
@@ -5,6 +5,7 @@ import logging
 import queue
 import threading
 import time
+import warnings
 from datetime import datetime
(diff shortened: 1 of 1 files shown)


hoh:
What about removing the old field completely instead ? I am okay with such breaking changes as long as the release notes are explicit about them.

MHHukiewitz:
> What about removing the old field completely instead ? I am okay with such breaking changes as long as the release notes are explicit about them.

I would prefer to phase it out later so that we can completely avoid any incompatibilities between SDK and pyaleph.

odesenfans (approved):

PR #33, opened 2023-07-05
Fix: SDK error when retrieving fallback private key, but the symlink was dead.

1yam:
Problem: The symlink can exist, but it can be dead. For example, unit tests may remove the file but not the symlink.

Solution: We add a check to determine if it is a symlink and if it is dead. In this case, we unlink it.

### DIFF
--- a/src/aleph/sdk/chains/common.py
+++ b/src/aleph/sdk/chains/common.py
@@ -120,7 +120,11 @@ def get_fallback_private_key(path: Optional[Path] = None) -> bytes:
         path.write_bytes(private_key)
 
         default_key_path = path.parent / "default.key"
+
+        # We check if the symlink exist but dead to remove it
+        if default_key_path.is_symlink() and not default_key_path.resolve().exists():
+            default_key_path.unlink()
+
+        # We create the symlink if symlink not exist
         if not default_key_path.exists():
-            # Create a symlink to use this key by default
             default_key_path.symlink_to(path)
-    return private_key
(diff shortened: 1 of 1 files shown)


src/aleph/sdk/chains/common.py
```diff
@@ -120,7 +120,12 @@ def get_fallback_private_key(path: Optional[Path] = None) -> bytes:
         path.write_bytes(private_key)
 
         default_key_path = path.parent / "default.key"
+
+        # We check if the symlink exist but dead to remove it
```
odesenfans, line 124:
```suggestion
        # If the symlink exists but does not point to a file, delete it.
```

src/aleph/sdk/chains/common.py
```diff
@@ -120,7 +120,12 @@ def get_fallback_private_key(path: Optional[Path] = None) -> bytes:
...
+        if default_key_path.is_symlink() and not default_key_path.resolve().exists():
+            default_key_path.unlink()
+
+        # We create the symlink if symlink not exist
```
odesenfans, line 128: Keep the old comment?
```suggestion
        # Create a symlink to use this key by default
```

odesenfans (changes requested):
Hablo ingles?

src/aleph/sdk/chains/common.py
```diff
@@ -120,7 +120,12 @@ def get_fallback_private_key(path: Optional[Path] = None) -> bytes:
         path.write_bytes(private_key)
 
         default_key_path = path.parent / "default.key"
+
+        # If the symlink exists but does not point to a file, delete it.
+        if default_key_path.is_symlink() and not default_key_path.resolve().exists():
+            default_key_path.unlink()
```
hoh, line 126: Can you log a warning in this case ? See the use of `logging` in other files.

1yam: Fix in a418aafd1ce416199721d3d5e43627b9fe62414b

odesenfans (approved):

PR #36, opened 2023-07-10
Feature: download on ipfs client

1yam:
Solutions:
   - Add a function called download_file_ipfs that downloads from ipfs.io/ipfs/.
   - Refactor the current download process to include:
      - Chunk size and storage into a buffer (BytesIO). (memory friendly)
   - Add a test for the download functionality.

### DIFF
--- a/src/aleph/sdk/client.py
+++ b/src/aleph/sdk/client.py
@@ -22,9 +22,9 @@ from typing import (
     Type,
     TypeVar,
     Union,
+    Protocol,
 )
 from io import BytesIO
-from typing import BinaryIO
 
 import aiohttp
 from aleph_message.models import (
@@ -69,6 +69,28 @@ except ImportError:
     magic = None  # type:ignore
 
 T = TypeVar("T")
+C = TypeVar("C", str, bytes, covariant=True)
+U = TypeVar("U", str, bytes, contravariant=True)
+
+
+class AsyncReadable(Protocol[C]):
+    async def read(self, n: int = -1) -> C:
+        ...
+
+
+class Writable(Protocol[U]):
+    def write(self, buffer: U) -> int:
+        ...
+
+
+async def copy_async_readable_to_buffer(
+    readable: AsyncReadable[C], buffer: Writable[C], chunk_size: int
+):
+    while True:
+        chunk = await readable.read(chunk_size)
+        if not chunk:
+            break
+        buffer.write(chunk)
 
 
 def async_wrapper(f):
@@ -231,6 +253,12 @@ class UserSessionSync:
 
     def download_file(self, file_hash: str) -> bytes:
         return self._wrap(self.async_session.download_file, file_hash=file_hash)
+
+    def download_file_ipfs(self, file_hash: str) -> bytes:
+        return self._wrap(
+            self.async_session.download_file_ipfs,
+            file_hash=file_hash,
+        )
 
     def watch_messages(
         self,
@@ -443,33 +471,6 @@ class AuthenticatedUserSessionSync(UserSessionSync):
         )
 
 
-async def download_file_to_buffer(
-    file_hash: str,
-    output_buffer: BinaryIO,
-) -> None:
-    """
-    Download a file from the storage engine and write it to the specified output buffer.
-
-    :param file_hash: The hash of the file to retrieve.
-    :param output_buffer: The binary output buffer to write the file data to.
-    """
-    url: str = f"{settings.API_HOST}/api/v0/storage/raw/{file_hash}"
-
-    ipfs_hash = ItemHash(file_hash)
-    if ItemType.from_hash(ipfs_hash) == ItemType.ipfs:
-        url = f"https://ipfs.aleph.im/ipfs/{file_hash}"
-    async with aiohttp.ClientSession() as session:
-        async with session.get(url) as response:
-            if response.status == 200:
-                while True:
-                    chunk = await response.content.read(16384)
-                    if not chunk:
-                        break
-                    output_buffer.write(chunk)
-            elif response.status == 413:
-                raise FileTooLarge(f"The file from {file_hash} is too large")
-
-
 class AlephClient:
     api_server: str
     http_session: aiohttp.ClientSession
@@ -639,6 +640,58 @@ class AlephClient:
             resp.raise_for_status()
             return await resp.json()
 
+    async def download_file_to_buffer(
+        self,
+        file_hash: str,
+        output_buffer: Writable[bytes],
+    ) -> None:
+        """
+        Download a file from the storage engine and write it to the specified output buffer.
+        :param file_hash: The hash of the file to retrieve.
+        :param output_buffer: Writable binary buffer. The file will be written to this buffer.
+        """
+
+        async with aiohttp.ClientSession() as session:
+            async with self.http_session.get(
+                f"/api/v0/storage/raw/{file_hash}"
+            ) as response:
+                if response.status == 200:
+                    await copy_async_readable_to_buffer(
+                        response.content, output_buffer, chunk_size=16 * 1024
+                    )
+                if response.status == 413:
+                    ipfs_hash = ItemHash(file_hash)
+                    if ItemType.from_hash(ipfs_hash) == ItemType.ipfs:
+                        return await self.download_file_ipfs_to_buffer(
+                            file_hash, output_buffer
+                        )
+                    else:
+                        raise FileTooLarge(f"The file from {file_hash} is too large")
+
+    async def download_file_ipfs_to_buffer(
+        self,
+        file_hash: str,
+        output_buffer: Writable[bytes],
+    ) -> None:
+        """
+        Download a file from the storage engine and write it to the specified output buffer.
+
+        :param file_hash: The hash of the file to retrieve.
+        :param output_buffer: The binary output buffer to write the file data to.
+        """
+        async with aiohttp.ClientSession() as session:
+            async with session.get(
+                f"https://ipfs.aleph.im/ipfs/{file_hash}"
+            ) as response:
+                if response.status == 200:
+                    await copy_async_readable_to_buffer(
+                        response.content, output_buffer, chunk_size=16 * 1024
+                    )
+                elif response.status == 413:
+                    raise FileTooLarge()
+                else:
+                    response.raise_for_status()
+
     async def download_file(
         self,
         file_hash: str,
@@ -651,7 +704,22 @@ class AlephClient:
         :param file_hash: The hash of the file to retrieve.
         """
         buffer = BytesIO()
-        await download_file_to_buffer(file_hash, output_buffer=buffer)
+        await self.download_file_to_buffer(file_hash, output_buffer=buffer)
+        return buffer.getvalue()
+
+    async def download_file_ipfs(
+        self,
+        file_hash: str,
+    ) -> bytes:
+        """
+        Get a file from the ipfs storage engine as raw bytes.
+
+        Warning: Downloading large files can be slow.
+
+        :param file_hash: The hash of the file to retrieve.
+        """
+        buffer = BytesIO()
+        await self.download_file_ipfs_to_buffer(file_hash, output_buffer=buffer)
         return buffer.getvalue()
 
     async def get_messages(
--- a/src/aleph/sdk/client.py
+++ b/src/aleph/sdk/client.py
@@ -650,7 +650,7 @@ class AlephClient:
         :param chunk_size: Size of chunk we download.
         """
         async with aiohttp.ClientSession() as session:
-            async with session.get(f"https://ipfs.io/ipfs/{file_hash}") as response:
+            async with session.get(f"https://ipfs.aleph.im/ipfs/{file_hash}") as response:
                 response.raise_for_status()
                 while True:
                     chunk = await response.content.read(chunk_size)
--- a/src/aleph/sdk/client.py
+++ b/src/aleph/sdk/client.py
@@ -55,6 +55,7 @@ from .exceptions import (
     InvalidMessageError,
     MessageNotFoundError,
     MultipleMessagesError,
+    FileTooLarge,
 )
 from .models import MessagesResponse
 from .utils import check_unix_socket_valid, get_message_type_value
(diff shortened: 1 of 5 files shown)


src/aleph/sdk/client.py
```diff
@@ -608,22 +613,85 @@ async def get_posts(
...
+    ) -> None:
+        """
+        Download a file from the storage engine and write it to the specified output buffer.
+
+        :param file_hash: The hash of the file to retrieve.
+        :param output_buffer: The binary output buffer to write the file data to.
+        :param chunk_size: Size of chunk we download.
+        """
+        async with aiohttp.ClientSession() as session:
+            async with session.get(f"https://ipfs.io/ipfs/{file_hash}") as response:
```
hoh, line 653: Can you use the IPFS server from aleph.im instead ?

src/aleph/sdk/client.py
```diff
@@ -608,22 +613,85 @@ async def get_posts(
...
+    ) -> None:
+        """
+        Download a file from the storage engine and write it to the specified output buffer.
+        :param self: The current instance of the class.
+        :param file_hash: The hash of the file to retrieve.
+        :param output_buffer: The binary output buffer to write the file data to.
+        :param chunk_size: Size of the chunk to download.
+        """
+        async with aiohttp.ClientSession() as session:
+            async with session.get(f"https://api1.aleph.im/api/v0/storage/raw/{file_hash}") as response:
```
hoh, line 630: Using our load balancer would help keeping this work when server api1 is under maintenance.
```suggestion
            async with session.get(f"https://official.aleph.im/api/v0/storage/raw/{file_hash}") as response:
```

There is already a setting in the SDK that points to the API server, `settings.API_HOST`.

src/aleph/sdk/client.py
hoh, line 653: Can you use the IPFS server from aleph.im instead ?
```suggestion
            async with session.get(f"https://ipfs.aleph.im/ipfs/{file_hash}") as response:
```

src/aleph/sdk/client.py
```diff
@@ -608,22 +613,85 @@ async def get_posts(
...
 
         :param file_hash: The hash of the file to retrieve.
+        :param chunk_size: The size of each chunk to read from the response.
         """
-        async with self.http_session.get(
-            f"/api/v0/storage/raw/{file_hash}"
-        ) as response:
-            response.raise_for_status()
-            return await response.read()
+        buffer = BytesIO()
```
hoh, line 674: Does this mean that everything is copied in memory ? What if the user downloads a 20 GB file ?

src/aleph/sdk/client.py
```diff
@@ -225,8 +227,11 @@ def get_posts(
             end_date=end_date,
         )
 
-    def download_file(self, file_hash: str) -> bytes:
-        return self._wrap(self.async_session.download_file, file_hash=file_hash)
+    def download_file(self, file_hash: str, chunk_size :int = 16 * 1024) -> bytes:
```
odesenfans, line 230: I see some formatting issues, apply black to your branch.

odesenfans: I don't think the chunk size needs to be part of the public API, it's an implementation detail. Implement it as a configuration variable or even as a constant in the code.

src/aleph/sdk/client.py
```diff
@@ -608,22 +613,85 @@ async def get_posts(
...
 
+    async def download_file_to_buffer(
+            self,
+            file_hash: str,
+            output_buffer: BinaryIO,
+            chunk_size: int,
+    ) -> None:
+        """
+        Download a file from the storage engine and write it to the specified output buffer.
+        :param self: The current instance of the class.
```
odesenfans, line 624: You can omit documentation for `self`.

src/aleph/sdk/client.py
odesenfans: Use `self.http_session`, we initialize one session when creating the `AlephClient` object.

src/aleph/sdk/client.py
odesenfans: This function returns the file as bytes, so there's little to be done about this issue. @1yam just preserved the public API, but I think we need to change the signature of this function to return `None` and just write to the file system. Same for `download_file_ipfs` below.

tests/unit/test_download.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+@pytest.mark.asyncio
+def test_download():
+    with AlephClient(
+        api_server=sdk_settings.API_HOST
+    ) as client:
+        file_size: int = 0
+        try:
+            file_content = client.download_file("QmeomffUNfmQy76CQGy9NdmqEnnHU9soCexBnGU3ezPHVH")  # File is 5B
```
odesenfans, line 13: Use the async version in your tests.

tests/unit/test_download.py
```diff
@@ -0,0 +1,32 @@
+import pytest
+from aleph.sdk import AlephClient
+from aleph.sdk.conf import settings as sdk_settings
+
+
+@pytest.mark.asyncio
+def test_download():
+    with AlephClient(
```
odesenfans, line 8: `async with`

tests/unit/test_download.py
```diff
@@ -0,0 +1,32 @@
...
+        api_server=sdk_settings.API_HOST
+    ) as client:
+        file_size: int = 0
+        try:
+            file_content = client.download_file("QmeomffUNfmQy76CQGy9NdmqEnnHU9soCexBnGU3ezPHVH")  # File is 5B
+            file_size = len(file_content)
+        except Exception as e:
```
odesenfans, line 15: No need to catch the exception here, you can let `pytest` catch the exception and log it.

tests/unit/test_download.py
```diff
@@ -0,0 +1,32 @@
...
+            pass
+        assert file_size == 5
```
odesenfans, line 17: Just compare to the expected content of the file, if it's 5 bytes you can just hardcode it.

odesenfans (changes requested):

src/aleph/sdk/client.py
hoh: Returning a [NamedTemporaryFile](https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile) would make more sense than returning `None`

hoh: https://github.com/aleph-im/aleph-sdk-python/issues/37

src/aleph/sdk/client.py
1yam: https://github.com/1yam/aleph-sdk-python/blob/cea3efb270ff6a5161a3a9f788e59e47ae9c1050/src/aleph/sdk/client.py#L661

src/aleph/sdk/client.py
1yam: fixed

src/aleph/sdk/client.py
1yam: fix

src/aleph/sdk/client.py
```diff
@@ -225,8 +228,17 @@ def get_posts(
             end_date=end_date,
         )
 
-    def download_file(self, file_hash: str) -> bytes:
-        return self._wrap(self.async_session.download_file, file_hash=file_hash)
+    def download_file(self, file_hash: str, chunk_size: int = 16 * 1024) -> bytes:
```
odesenfans, line 231: I still don't think the chunk size should appear in public functions.

src/aleph/sdk/client.py
```diff
@@ -608,22 +620,100 @@ async def get_posts(
...
+        output_buffer: BinaryIO,
+        chunk_size: int,
+    ) -> None:
+        """
+        Download a file from the storage engine and write it to the specified output buffer.
+        :param file_hash: The hash of the file to retrieve.
+        :param output_buffer: Writable binary buffer. The file will be written to this buffer.
+        :param chunk_size: Size of the chunk to download.
+        """
+        IPFS_HASH = ItemHash(file_hash)
```
odesenfans, line 635: Why upper case? Also you don't need this variable if the file is downloaded through the API. Move it in the if 413 condition..

src/aleph/sdk/client.py
```diff
@@ -608,22 +620,100 @@ async def get_posts(
...
+                        if not chunk:
+                            break
+                        output_buffer.write(chunk)
+                if response.status == 413:
+                    if ItemType.from_hash(IPFS_HASH) == ItemType.ipfs:
+                        return await self.download_file_ipfs_to_buffer(
+                            file_hash, output_buffer, chunk_size
+                        )
+                    else:
+                        raise Exception("Unsupported file hash")
```
odesenfans, line 654: Raising `Exception` is too broad because it makes it hard for users to determine the nature of the problem. Here the problem is not that we do not support other hash types, it's just that the file is too big to be downloaded through the API. What you should do is create a `FileTooLarge` exception in the `exceptions.py` module and raise it here.

src/aleph/sdk/client.py
```diff
@@ -608,22 +620,100 @@ async def get_posts(
...
+        :param chunk_size: Size of the chunk to download.
+        """
+        IPFS_HASH = ItemHash(file_hash)
+
+        async with aiohttp.ClientSession() as session:
+            async with self.http_session.get(
+                f"/api/v0/storage/raw/{file_hash}"
+            ) as response:
+                if response.status == 200:
+                    response.raise_for_status()
```
odesenfans, line 642: This is useless, no need to `raise_for_status()` if you know that the HTTP code is 200. `raise_for_status()` only raises an exception for 4xx/5xx HTTP codes.

src/aleph/sdk/client.py
```diff
@@ -608,22 +620,100 @@ async def get_posts(
...
+                    while True:
+                        chunk = await response.content.read(chunk_size)
+                        if not chunk:
+                            break
+                        output_buffer.write(chunk)
```
odesenfans, line 647: This code is duplicated when reading from IPFS as well. Could you encapsulate it in a function?

Ex:

```
from typing import Protocol, TypeVar

T = TypeVar("T", bytes, str)


class AsyncReadable(Protocol[T]):
  async def read(n: int = -1) -> T:
    ...

class Writable(Protocol[T]):
  def write(buffer: T) -> int:
    ...

async def copy_async_readable_to_buffer(readable: AsyncReadable[T], buffer: Writable[T], chunk_size: int):
  ...
```

odesenfans (changes requested):

src/aleph/sdk/client.py
```diff
@@ -439,6 +443,33 @@ def submit(
...
+    async with aiohttp.ClientSession() as session:
+        async with session.get(url) as response:
+            if response.status == 200:
+                while True:
+                    chunk = await response.content.read(16384)
+                    if not chunk:
+                        break
+                    output_buffer.write(chunk)
+            elif response.status == 413:
+                raise FileTooLarge(f"The file from {file_hash} is too large")
```
odesenfans, line 470: The logic is still not what's needed. Here you replace the URL directly to go to the IPFS service, which is not what we want. I want to go by default to the aleph.im network and then fall back to the IPFS service if we encounter a 413. I don't want to overload the IPFS service for small files.

src/aleph/sdk/client.py
```diff
@@ -439,6 +443,33 @@ def submit(
         )
 
 
+async def download_file_to_buffer(
```
odesenfans, line 446: Why did you make it a function instead of a method? I think this function is quite convenient to keep inside of `AlephClient`.

src/aleph/sdk/exceptions.py
```diff
@@ -42,3 +42,11 @@ class BadSignatureError(Exception):
     """
 
     pass
+
+
+class FileTooLarge(Exception):
+    """
+    A file is too large
```
odesenfans, line 49: Fair enough.

odesenfans (changes requested):

src/aleph/sdk/client.py
```diff
@@ -65,6 +69,28 @@
     magic = None  # type:ignore
 
 T = TypeVar("T")
+C = TypeVar("C", str, bytes, covariant=True)
```
odesenfans, line 72: You can simply use `T = TypeVar("T", str, bytes, covariant=True)`, it's supposed to work.

src/aleph/sdk/client.py
```diff
@@ -65,6 +69,28 @@
...
+
+
+async def copy_async_readable_to_buffer(
+    readable: AsyncReadable[C], buffer: Writable[C], chunk_size: int
+):
+    while True:
+        chunk = await readable.read(chunk_size)
+        if not chunk:
+            break
+        buffer.write(chunk)
```
odesenfans, line 93: Also, maybe move the implementation of this function in a new module, like `aleph.sdk.toolkit.io`.

src/aleph/sdk/client.py
```diff
@@ -608,6 +640,58 @@ async def get_posts(
...
+        file_hash: str,
+        output_buffer: Writable[bytes],
+    ) -> None:
+        """
+        Download a file from the storage engine and write it to the specified output buffer.
+        :param file_hash: The hash of the file to retrieve.
+        :param output_buffer: Writable binary buffer. The file will be written to this buffer.
+        """
+
+        async with aiohttp.ClientSession() as session:
```
odesenfans, line 654: This session is not used.

src/aleph/sdk/client.py
```diff
@@ -608,6 +640,58 @@ async def get_posts(
...
+            async with self.http_session.get(
+                f"/api/v0/storage/raw/{file_hash}"
+            ) as response:
+                if response.status == 200:
+                    await copy_async_readable_to_buffer(
+                        response.content, output_buffer, chunk_size=16 * 1024
+                    )
+                if response.status == 413:
+                    ipfs_hash = ItemHash(file_hash)
+                    if ItemType.from_hash(ipfs_hash) == ItemType.ipfs:
```
odesenfans, line 664: You can directly use `ipfs_hash.item_type`.

src/aleph/sdk/client.py
```diff
@@ -608,6 +640,58 @@ async def get_posts(
...
+        """
+        async with aiohttp.ClientSession() as session:
+            async with session.get(
+                f"https://ipfs.aleph.im/ipfs/{file_hash}"
+            ) as response:
+                if response.status == 200:
+                    await copy_async_readable_to_buffer(
+                        response.content, output_buffer, chunk_size=16 * 1024
+                    )
+                elif response.status == 413:
```
odesenfans, line 690: This call will not return a 413, remove this branch.

odesenfans (changes requested):

odesenfans (dismissed):

odesenfans (approved):

PR #38, opened 2023-07-10
Fix: VmCache did not call the API via unix socket

hoh:
Instead, it tried using TCP calls on localhost:80.

Solution: Detect if `settings.API_UNIX_SOCKET` is set and use it when available.

Related to https://github.com/aleph-im/aleph-vm/pull/368/


odesenfans:
This PR is not on top of my changes last week. Pretty sure this is a duplicate of #35.

PR #51, opened 2023-08-19
Add support for signing messages using LedgerHQ wallet on Ethereum

hoh:
The ability to use private keys stored on a hardware wallet to sign messages would greatly enhance the security of the network, especially since messages can be amended and the virtual machines will potentially have many side effects.

The command / response scheme used by LedgerHQ to address the device is similar to the ISO/IEC 7816-4 smartcard protocol. Each command / response packet is called an APDU (application protocol data unit).

Each APDU is specific to Ledger application, that adds the support for a chain or functionality.

The library [ledgereth](https://pypi.org/project/ledgereth/) implements the LedgerHQ ADPU for the Ethereum application.

This branch adds support in the SDK.

Usage:

```python
>>> import asyncio
>>> from aleph.sdk.client import AuthenticatedAlephClient
>>> from aleph.sdk.wallets.ledger.ethereum import LedgerETHAccount
>>> account = LedgerETHAccount.from_address("0x...")
>>> client = AuthenticatedAlephClient(account=account, api_server="https://api2.aleph.im")
>>> await client.create_post(post_content="coucou", post_type="ALEPH_TEST")
```

Fixes https://github.com/aleph-im/aleph-client/issues/13


src/aleph/sdk/wallets/ledger/ethereum.py
```diff
@@ -0,0 +1,74 @@
...
+
+class LedgerETHAccount(BaseAccount):
+    """Account using the Ethereum app on Ledger hardware wallets."""
+
+    CHAIN = "ETH"
+    CURVE = "secp256k1"
+    _account: LedgerAccount
+    _device: Dongle
+
+    def __init__(self, account: LedgerAccount, device: Optional[Dongle]):
```
odesenfans, line 23: Do you want the user to create the object using `from_address`, `__init__` or both? Seems like there is a bit of duplication when calling `init_dongle()`.

src/aleph/sdk/wallets/ledger/ethereum.py
```diff
@@ -0,0 +1,74 @@
...
+from ledgerblue.Dongle import Dongle
+from ledgereth import find_account, get_account_by_path, get_accounts
+from ledgereth.comms import init_dongle
+from ledgereth.messages import sign_message
+from ledgereth.objects import LedgerAccount, SignedMessage
+
+from ...chains.common import BaseAccount, get_verification_buffer
+
+
+class LedgerETHAccount(BaseAccount):
```
odesenfans, line 15: Wouldn't it make sense to create a `LedgerAccount` base class that's customizable for each chain afterwards?

tests/unit/test_wallet_ethereum.py
```diff
@@ -0,0 +1,47 @@
...
+class Message:
+    chain: str
+    sender: str
+    type: str
+    item_hash: str
+
+
+@pytest.mark.ledger_hardware
+@pytest.mark.asyncio
+async def test_LedgerETHAccount():
```
odesenfans, line 21: Nitpick: snake case?

```suggestion
async def test_ledger_eth_account():
```

tests/unit/test_wallet_ethereum.py
```diff
@@ -0,0 +1,47 @@
...
+    )
+
+    with pytest.raises(BadSignatureError):
+        signed["signature"] = signed["signature"][:-8] + "cafecafe"
+
+        verify_signature(
+            signed["signature"], signed["sender"], get_verification_buffer(signed)
+        )
+
+    # Obtaining the public key is not supported (yet ?) on hardware wallets.
```
odesenfans, line 45: Really? Could you add a link to some documentation in the implementation of `get_public_key` then?

tests/unit/test_wallet_ethereum.py
odesenfans, line 21: How can I run this test locally? Does it require root access to open the USB device?

odesenfans (changes requested):

src/aleph/sdk/wallets/ledger/ethereum.py
hoh: I removed the `init_dongle()` from the `__init__` and added docstrings.

src/aleph/sdk/wallets/ledger/ethereum.py
hoh: Only once we get support for other chains, creating an empty, unused class would not make sense.

In order to support other chains, we will need to decide on the best strategy since [ledgereth](https://pypi.org/project/ledgereth/) only supports Ethereum, but seems to contain most of the boiler plate for supporting other chains as well.

tests/unit/test_wallet_ethereum.py
hoh: Moved to a docsting in the implementation.

PR #57, opened 2023-08-28
Feature: authenticated file upload SDK

1yam:
SDK implementations of  : https://github.com/aleph-im/pyaleph/pull/463


examples/store.py
```diff
@@ -3,6 +3,7 @@
 import click
 from aleph_message.models import StoreMessage
 from aleph_message.status import MessageStatus
+from mypy.dmypy_server import MiB
```
hoh, line 6: From `mypy` ?

examples/store.py
```diff
@@ -54,6 +55,30 @@ async def do_upload(account, engine, channel, filename=None, file_hash=None):
         await print_output_hash(message, status)
 
 
+async def do_upload_with_message(account, engine, channel, filename=None, item_hash=None):
```
hoh, line 58: Can you add type annotations ?
```suggestion
async def do_upload_with_message(account: ..., engine: ..., channel: str, filename: Optional[Path]=None, item_hash: Optional[ItemHash]=None):
```

examples/store.py
```diff
@@ -54,6 +55,30 @@ async def do_upload(account, engine, channel, filename=None, file_hash=None):
...
+    async with AuthenticatedAlephClient(
+        account=account, api_server=settings.API_HOST
+    ) as session:
+        print(filename, account.get_address())
+        if filename:
+            try:
+                with open(filename, "rb") as f:
```
hoh, line 65: If you are to put everything in memory during the upload, then `https://docs.python.org/3/library/pathlib.html#pathlib.Path.read_bytes` would simplify the code.

However it is a good practice to not put the file content in memory and instead use streaming upload: https://docs.aiohttp.org/en/stable/client_quickstart.html?highlight=upload#streaming-uploads

MHHukiewitz (dismissed):

examples/store.py
MHHukiewitz: This is the only suggestion that has not been implemented *ideally*, `read_bytes` is being used now. I'll leave the decision to merge with @hoh.

PR #54, opened 2023-08-28
AlephClientBase and AuthenticatedAlephClientBase

MHHukiewitz:
Adding abstract base classes for the client classes and add several improvements:
- Many pages of messages can now be iterated over with `get_messages_iterator` or `get_posts_iterator`.
- `get_posts` now returns a `PostsResponse` type instead of untyped dictionary
- Harmonize `get_messages` and `watch_messages` as both should have the same query parameters
- Moved all docstrings from the client classes to their ABCs

PR #58, opened 2023-08-28
Refactor: Filters were messy with duplication

hoh:
The management of query filters on API requests was coded inside other functions, hard to read and difficult to read.

Solution: Rely on a `MessageQueryFilter` object that manages the filters.

A new `MessageQuery` interface allows an easier handling of pagination, including iterating transparently iterating over messages with `async for message in MessageQuery(query_filter=filter)`.

The interface of `AlephClient.get_messages` has been kept unchanged for now for compatibility reasons.

Inspired by #54 from @MHHukiewitz

### DIFF
--- a/src/aleph/sdk/client.py
+++ b/src/aleph/sdk/client.py
@@ -45,7 +45,6 @@ from aleph_message.models import (
 )
 from aleph_message.models.execution.base import Encoding
 from aleph_message.status import MessageStatus
-from pydantic import ValidationError
 
 from aleph.sdk.types import Account, GenericMessage, StorageEnum
 from aleph.sdk.utils import Writable, copy_async_readable_to_buffer
@@ -59,6 +58,7 @@ from .exceptions import (
     MultipleMessagesError,
 )
 from .models import MessagesResponse
+from .query import MessageQuery, MessageQueryFilter
 from .utils import check_unix_socket_valid, get_message_type_value
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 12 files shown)


src/aleph/sdk/query.py
```diff
@@ -0,0 +1,211 @@
...
+        Handles pagination internally.
+
+        ```
+        async for message in MessageQuery(query_filter=filter):
+            print(message)
+        ```
+        """
+        page: int = 0
+        partial_result = await self.fetch(page=0)
+        while partial_result:
```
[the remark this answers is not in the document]
MHHukiewitz: Even if not, it will result in one additional request to the API, which could've been prevented by comparing `pagination_total` with `pagination_page * pagination_per_page`.

src/aleph/sdk/client.py
```diff
@@ -736,96 +736,28 @@ async def get_messages(
...
-                pagination_total=response_json["pagination_total"],
-                pagination_per_page=response_json["pagination_per_page"],
-                pagination_item=response_json["pagination_item"],
-            )
+        return await MessageQuery(
+            query_filter=query_filter,
+            http_client_session=self.http_session,
+            ignore_invalid_messages=ignore_invalid_messages,
+            invalid_messages_log_level=invalid_messages_log_level,
+        ).fetch(page=page, pagination=pagination)
```
MHHukiewitz, line 760: Seeing this, I wonder if we should ditch the `.fetch(page=page, pagination=pagination)` call to allow users to decide whether to fetch a certain page or to iterate through all the messages?

Currently I cannot see how a user would actually access the iterator of `MessageQuery`.

MHHukiewitz (commented):
Highly appreciated changes! Left a few comments that highlight some issues that might appear with the message iterator.

src/aleph/sdk/client.py
MHHukiewitz: Regarding this, I cannot help but feel reminded of the way I implemented similar functionality on the Active Record SDK, see: https://github.com/aleph-im/active-record-sdk/blob/main/src/aars/utils.py
 
Here, we have `IndexQuery` which prepares the parameters like `MessageQueryFilter` does.
The `AARS` client class handles the actual requests, which is here being handled by `MessageQuery`.
Then `PageableRequest` & `PageableResponse` give all the pagination & iteration over the response of the executed `MessageQuery`.

I got to admit though, that your approach looks more elegant than what I did in the Active Record SDK, but some of this uglyness resulted in the fact that:

1. I am using an ORM and all the functionality needs to be attached to the base class of `Records` (which actually represent POSTs).
2. Some responses are async generators, while some are sync, as AARS does not have a cache capable of async iteration/pagination (while my DomainNode implementation in #59 does).

I think there is some significant overlap between this PR, the PRs #54 and #59 and the Active Record SDK. Given the right preparation and time for thought, we could build a more coherent and powerful SDK for handling messages as well as POSTs.

One more thing about POSTs: Given the interesting behavior that is possible by amending these messages, I think that they would benefit from the ORM approach chosen in AARS, as it would give users very simple syntax to update/amend POSTs as well as retrieving past revisions of them. This is a paradigm not often seen in DLTs and should be highlighted and easier to work with, IMHO.

hoh:
@MHHukiewitz is this PR deprecated by #65 ?

MHHukiewitz:
@hoh yes, it is. I am making an issue out of my last comment, so we can close hits

PR #59, opened 2023-09-05
MessageCache and LightNode

MHHukiewitz:
Reopened with smaller scope from #31. Depends on #54.

## Problem
Some apps require access to the same messages over and over again. This puts unnecessary strain on the CCN API and increases the latency of apps using Aleph Messages.

## Solution
This pull request proposes a solution by introducing a message cache implemented with Peewee and an SQLite database in the Aleph SDK. The message cache effectively stores frequently accessed messages, eliminating the need for repetitive CCN API calls. With the message cache in place, apps utilizing the Aleph SDK will experience enhanced efficiency and responsiveness in accessing messages.


MHHukiewitz:
I started using v1 posts instead of v0 posts. I wonder why we did not earlier, as it seems a cleaner, albeit minimalistic, endpoint for getting posts. There should be no problem in changing back to v0 posts though.

MHHukiewitz:
A few points on my ToDo list:

- `DomainNode`: Should keep a synchronized cache with a subset of all Aleph messages.
  - Only allow messages to be posted, that fall under the scope of the _Domain_, meaning they would not be filtered by given initialization parameters. 
  - Should remove messages from the cache, if they not reappear as `confirmed` in the watch_messages stream, after some pre-defined timeout.
- Make `BaseAlephClient` and `BaseAuthenticatedAlephClient` pure interfaces.
- Add interfaces for `UserSessionSync` and `AuthenticatedUserSessionSync`
  - Let `MessageCache` implement `UserSessionSyncInterface` instead of `BaseAlephClient`, as all provided methods can be sync.
  - Implement `DomainNode.__enter__()` with a `DomainNodeSync` class.

src/aleph/sdk/node/__init__.py
```diff
@@ -0,0 +1,663 @@
...
+            if not PostModel.table_exists():
+                db.create_tables([PostModel])
+
+        MessageCache._instance_count += 1
+
+    def __del__(self):
+        MessageCache._instance_count -= 1
+
+        if MessageCache._instance_count == 0:
+            db.close()
```
odesenfans, line 62: Looking at this logic, I'm thinking that a context manager may help. Don't know where to put it yet at this stage of the review though.

src/aleph/sdk/node/__init__.py
```diff
@@ -0,0 +1,663 @@
...
+        for item in iter(MessageModel.select().order_by(-MessageModel.time)):
+            yield model_to_message(item)
+
+    def __repr__(self) -> str:
+        return f"<MessageCache: {db}>"
+
+    def __str__(self) -> str:
+        return repr(self)
+
+    def add(self, messages: Union[AlephMessage, Iterable[AlephMessage]]):
```
odesenfans, line 97: I think this function could be reorganized for clarity.

1. Avoid looping twice, loop over all messages to get messages, posts and amends.
2. Insert the objects in the DB.
3. Hide the complex bits in sub-functions.

src/aleph/sdk/node/__init__.py
```diff
@@ -0,0 +1,663 @@
...
+            original_post.address = message.sender
+            original_post.channel = message.channel
+            original_post.last_updated = datetime.fromtimestamp(message.time)
+            post_data.append(original_post)
+
+        PostModel.insert_many(post_data).on_conflict_replace().execute()
+
+    def get(
+        self, item_hashes: Union[Union[ItemHash, str], Iterable[Union[ItemHash, str]]]
+    ) -> List[AlephMessage]:
```
odesenfans, line 152: Note that with typing overloads you can do:

```
@overload
def get(self, item_hash: ItemHash) -> Optional[AlephMessage]:
  ...
  
@overload
def get(self, item_hashes: Iterable[ItemHash]) -> List[AlephMessage]:
  ...

def get(self, item_hashes):
  # Implement here
  ...
```

src/aleph/sdk/node/__init__.py
```diff
@@ -0,0 +1,663 @@
...
+            original_post.original_type = message.content.type
+            original_post.address = message.sender
+            original_post.channel = message.channel
+            original_post.last_updated = datetime.fromtimestamp(message.time)
+            post_data.append(original_post)
+
+        PostModel.insert_many(post_data).on_conflict_replace().execute()
+
+    def get(
+        self, item_hashes: Union[Union[ItemHash, str], Iterable[Union[ItemHash, str]]]
```
odesenfans, line 151: Force the user to pass `ItemHash` objects.

src/aleph/sdk/node/__init__.py
```diff
@@ -0,0 +1,663 @@
...
+        async def _listen():
+            async for message in message_stream:
+                self.add(message)
+                print(f"Added message {message.item_hash} to cache")
+
+        return _listen()
+
+    async def fetch_aggregate(
+        self, address: str, key: str, limit: int = 100
+    ) -> Dict[str, Dict]:
```
odesenfans, line 180: Be careful about the meaning of the `limit` keyword here. It used to mean "the number of messages to aggregate" for performance reasons, and is now unused because we materialize the aggregates. if you want to implement pagination, using the `page` and `pagination` keywords like we do on other methods seems more appropriate.

odesenfans: Also, this function clearly does not perform aggregation?

src/aleph/sdk/node/__init__.py
```diff
@@ -0,0 +1,663 @@
...
+        return PostsResponse(
+            posts=posts,
+            pagination_page=page,
+            pagination_per_page=pagination,
+            pagination_total=query.count(),
+            pagination_item="posts",
+        )
+
+    async def download_file(self, file_hash: str) -> bytes:
+        raise NotImplementedError
```
odesenfans, line 246: Why is it not implemented?

src/aleph/sdk/node/__init__.py
```diff
@@ -0,0 +1,663 @@
...
+        Get a single message from the cache.
+        """
+        query = MessageModel.select().where(MessageModel.item_hash == item_hash)
+
+        if message_type:
+            query = query.where(MessageModel.type == message_type.value)
+        if channel:
+            query = query.where(MessageModel.channel == channel)
+
+        item = query.first()
```
odesenfans, line 313: Does peewee provide a `one()` method instead of `first()`? In SQLAlchemy the two are different, `one()` will raise an exception if several rows are returned.

src/aleph/sdk/node/__init__.py
```diff
@@ -0,0 +1,663 @@
...
+            end_date=end_date,
+        )
+
+        async for item in query:
+            yield model_to_message(item)
+
+
+class DomainNode(MessageCache, BaseAuthenticatedAlephClient):
+    """
+    A Domain Node is a queryable proxy for Aleph Messages that are stored in a database cache and/or in the Aleph
```
odesenfans, line 359: Documentation-wise, we should mention Aleph as "Aleph.im" when starting a sentence and "aleph.im" when inside a sentence.

src/aleph/sdk/node/__init__.py
```diff
@@ -0,0 +1,663 @@
...
+        )
+        self.chains = (
+            list(chains) + [Chain(session.account.CHAIN)]
+            if chains
+            else [session.account.CHAIN]
+        )
+        self.message_types = message_types
+
+        # start listening to the network and storing messages in the cache
+        asyncio.get_event_loop().create_task(
```
odesenfans, line 393: Can't you just call `asyncio.create_task()`? Using `get_event_loop` is considered a low-level primitive.

src/aleph/sdk/node/__init__.py
```diff
@@ -0,0 +1,663 @@
...
+                    tags=self.tags,
+                    addresses=self.addresses,
+                    chains=self.chains,
+                    message_types=self.message_types,
+                )
+            )
+        )
+
+        # synchronize with past messages
+        asyncio.get_event_loop().run_until_complete(
```
odesenfans, line 406: Same, use `asyncio.run()`.

src/aleph/sdk/node/__init__.py
```diff
@@ -0,0 +1,663 @@
...
+
+    async def download_file(self, file_hash: str) -> bytes:
+        """
+        Opens a file that has been locally stored by its hash.
+        """
+        try:
+            with open(self._file_path(file_hash), "rb") as f:
+                return f.read()
+        except FileNotFoundError:
+            file = await self.session.download_file(file_hash)
```
odesenfans, line 461: What if the file does not exist? Also, consider using a `pass` clause to get out of the `except` condition, otherwise you'll have tracebacks looking like "Another exception occurred while handling exception FileNotFoundError".

src/aleph/sdk/node/__init__.py
```diff
@@ -0,0 +1,663 @@
...
+            post_content=post_content,
+            post_type=post_type,
+            ref=ref,
+            address=address,
+            channel=channel,
+            inline=inline,
+            storage_engine=storage_engine,
+            sync=sync,
+        )
+        # WARNING: this can cause inconsistencies if the message is dropped/rejected by the aleph node
```
odesenfans, line 521: To resolve this you can add a background task that waits for the processing of all pending messages and adds them to the cache if they are processed successfully.

src/aleph/sdk/node/common.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+class JSONDictEncoder(json.JSONEncoder):
+    def default(self, obj):
+        if isinstance(obj, BaseModel):
+            return obj.dict()
+        return json.JSONEncoder.default(self, obj)
+
+
+pydantic_json_dumps = partial(json.dumps, cls=JSONDictEncoder)
```
odesenfans, line 22: You can just use

```
from pydantic.json import pydantic_encoder

pydantic_json_dumps = partial(json.dumps, default=pydantic_encoder)
```

tests/integration/itest_forget.py
```diff
@@ -100,31 +100,5 @@ async def test_forget_a_forget_message(fixture_account):
...
-
-        get_forget_message_response = await session.get_messages(
-            hashes=[forget_message_hash],
-            channels=[TEST_CHANNEL],
-        )
-        assert len(get_forget_message_response.messages) == 1
-        forget_message = get_forget_message_response.messages[0]
-        print(forget_message)
-
-        assert "forgotten_by" not in forget_message
```
odesenfans, line 130: Why is this removed in this PR?

odesenfans (dismissed):
Reviewed ~70% of it, some changes and questions.

src/aleph/sdk/node/__init__.py
MHHukiewitz: Would make sense to implement, indeed.

src/aleph/sdk/node/__init__.py
MHHukiewitz: Nope, it's the only way: https://docs.peewee-orm.com/en/latest/peewee/querying.html#selecting-a-single-record

But I'm taking care of it in `get_message()`.

src/aleph/sdk/node/__init__.py
MHHukiewitz: The second loop is there to ensure correct behavior when receiving unordered messages. Could this occur in the wild?

src/aleph/sdk/node/__init__.py
```diff
@@ -0,0 +1,663 @@
...
+    missing_posts: Dict[ItemHash, PostMessage] = {}
+    """A dict of all posts by item_hash and their amend messages that are missing from the cache."""
+
+    def __init__(self):
+        if db.is_closed():
+            db.connect()
+            if not MessageModel.table_exists():
+                db.create_tables([MessageModel])
+            if not PostModel.table_exists():
+                db.create_tables([PostModel])
```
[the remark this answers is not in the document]
MHHukiewitz: Right now, I'd think people would have to throw away their DB and resynchronize. Maybe I should add a check, whether the model is different from the actual table?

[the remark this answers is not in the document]
MHHukiewitz: I added a check that will raise an error and inform the user, if the schema seems to change.

MHHukiewitz:
Implemented all of the suggested changes by @odesenfans, highlights:
- Broadcasted messages are now being polled up to 50 seconds for their status every 5 seconds. If they fail to become `processed`, they are removed from the cache.
- Validating that the database schema is the same as defined in the SDK's model, raising an error and asking the user to delete or migrate the existing DB.
- Aggregates are now correctly being aggregated and materialized in an Aggregate table.
- Forget messages are now taken into consideration and they delete entries in the cache.
- File downloads are now cached, too.

MHHukiewitz:
> Failed to retrieve llama text: ('Connection aborted.', RemoteDisconnected('Remote end closed connection without response'))

Context size not enough, I rate this PR manually as `BLACK`

nesitor (approved):

hoh:
This PR adds +1,937 and -16 lines of code to the project.

What is the impact of this PR on the test coverage ?

src/aleph/sdk/__init__.py
```diff
@@ -1,6 +1,6 @@
 from pkg_resources import DistributionNotFound, get_distribution
 
-from aleph.sdk.client import AlephHttpClient, AuthenticatedAlephHttpClient
+from aleph.sdk.client import AlephHttpClient, AuthenticatedAlephHttpClient, LightNode
```
Psycojoker, line 3: Current discussion with @hoh : we might want to use another name for this class

PR #60, opened 2023-09-13
Fix: support 405 from POST /messages in broadcast fallback

odesenfans:
Problem: a user reports that publishing a message falls because of a 405 error returned from `POST /messages`. The fallback only takes 404 into account, but a 405 is actually more likely because `GET /messages` exists.

Solution: accept both 404 and 405 as a sign that the node does not support `POST /messages`.


nesitor (approved):

hoh:
From the spec:
> The server must generate an Allow header field in a 405 status code response. The field must contain a list of methods that the target resource currently supports.

Is this the case on pyaleph @odesenfans ?

Source: https://developer.mozilla.org/en-US/docs/web/http/status/405

PR #61, opened 2023-09-22
Fix: allow user to specify datetime and UUID objects in post content

odesenfans:
Problem: using `create_post()` with an object containing datetime, UUIDs and other types not serializable by the Python `json` module fails.

Solution: as we already use Pydantic all over the place, use the Pydantic encoder as default for all the types not serializable by the `json` module.


hoh:
Looks like a decent idea :+1:

hoh (approved):

PR #63, opened 2023-09-28
Refactor: Add MessageFilter & PostFilter

MHHukiewitz:
Depending on #59.

- Adding `MessageFilter` & `PostFilter` as suggested in #58.
- Fixes to `MessageCache` & `DomainNode` (need to be moved to #59)
- Refactored cache-related functions and move to new `models` & `models.db` modules.

PR #65, opened 2023-10-05
Refactor AlephClient classes

MHHukiewitz:
This PR is in preparation for the alpha release of v0.8.0 and contains all changes to the client's interface that are intended for the breaking release v0.8.0.

- Introduce `MessageFilter` and `PostFilter`, inspired by #58
- Remove obsolete `limit` parameter in `fetch_aggregate()` and `fetch_aggregates`
- Rename occurences of "Aleph" to "Aleph.im/aleph.im", inspired by @odesenfans
- Fix broken integration tests


src/aleph/sdk/base.py
```diff
@@ -26,42 +25,33 @@
...
 DEFAULT_PAGE_SIZE = 200
 
 
 class BaseAlephClient(ABC):
     @abstractmethod
-    async def fetch_aggregate(
-        self,
-        address: str,
-        key: str,
-        limit: int = 100,
```
hoh, line 41: Why remove the field `limit` ?

src/aleph/sdk/base.py
```diff
@@ -70,15 +60,7 @@ async def get_posts(
         self,
         pagination: int = DEFAULT_PAGE_SIZE,
```
hoh, line 61: While breaking the API of this method anyways, why not rename this `page_size` ?

src/aleph/sdk/base.py
```diff
@@ -318,7 +194,7 @@ async def create_post(
         sync: bool = False,
     ) -> Tuple[AlephMessage, MessageStatus]:
         """
-        Create a POST message on the Aleph network. It is associated with a channel and owned by an account.
+        Create a POST message on the aleph.im network. It is associated with a channel and owned by an account.
```
hoh, line 197: How is this commit related to this PR ? It would be better in a different PR for easier review of both.

src/aleph/sdk/client.py
```diff
@@ -591,31 +514,11 @@ async def get_posts(
...
-        if start_date is not None:
-            if not isinstance(start_date, float) and hasattr(start_date, "timestamp"):
-                start_date = start_date.timestamp()
-            params["startDate"] = start_date
-        if end_date is not None:
-            if not isinstance(end_date, float) and hasattr(start_date, "timestamp"):
-                end_date = end_date.timestamp()
-            params["endDate"] = end_date
+        if not post_filter:
+            post_filter = PostFilter()
```
hoh, line 518:
```suggestion
        post_filter = post_filter or PostFilter()
```

src/aleph/sdk/client.py
```diff
@@ -591,31 +514,11 @@ async def get_posts(
...
+        params = post_filter.as_http_params()
+        params["page"] = str(page)
+        params["pagination"] = str(pagination)
```
hoh, line 521:
```suggestion
        params = post_filter.as_http_params() | {
            "page" = str(page),
            "pagination" = str(pagination),
        }
```

src/aleph/sdk/client.py
```diff
@@ -746,43 +638,11 @@ async def get_messages(
...
-        if end_date is not None:
-            if not isinstance(end_date, float) and hasattr(start_date, "timestamp"):
-                end_date = end_date.timestamp()
-            params["endDate"] = end_date
-
+        if not message_filter:
+            message_filter = MessageFilter()
+        params = message_filter.as_http_params()
+        params["page"] = str(page)
+        params["pagination"] = str(pagination)
```
hoh, line 645: Sees comment on similar code

src/aleph/sdk/client.py
```diff
@@ -846,54 +708,11 @@ async def get_message(
...
-        if start_date is not None:
-            if not isinstance(start_date, float) and hasattr(start_date, "timestamp"):
-                start_date = start_date.timestamp()
-            params["startDate"] = start_date
-        if end_date is not None:
-            if not isinstance(end_date, float) and hasattr(start_date, "timestamp"):
-                end_date = end_date.timestamp()
-            params["endDate"] = end_date
+        if not message_filter:
+            message_filter = MessageFilter()
```
hoh, line 714:
```suggestion
        message_filter = message_filter or MessageFilter()
```

src/aleph/sdk/client.py
```diff
@@ -846,54 +708,11 @@ async def get_message(
...
+        params = message_filter.as_http_params()
```
hoh, line 715: What's the point of creating an empty `MessageFilter` is the goal is to `.as_http_params()` that will produce `""` ?

src/aleph/sdk/base.py
```diff
@@ -2,7 +2,6 @@
```
hoh, line 2: Why is this named `base.py` while it is the base of the client only ? What about `aleph.sdk.client.base` ?

src/aleph/sdk/models/common.py
```diff
@@ -0,0 +1,29 @@
...
+
+def _date_field_to_float(date: Optional[Union[datetime, float]]) -> Optional[float]:
+    if date is None:
+        return None
+    elif isinstance(date, float):
+        return date
+    elif hasattr(date, "timestamp"):
+        return date.timestamp()
+    else:
+        raise TypeError(f"Invalid type: `{type(date)}`")
```
hoh, line 29: These are not models, why are they not in an utility file instead ? Also, no docstring explaining why we need them ?

src/aleph/sdk/models/message.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+class MessagesResponse(PaginationResponse):
+    """Response from an aleph.im node API on the path /api/v0/messages.json"""
+
+    messages: List[AlephMessage]
+    pagination_item = "messages"
+
+
+class MessageFilter:
```
hoh, line 16: What is the point of this general `sdk.models` package ? It the goal is to put all classes there, it does not seem to make much sense to me.

This class is related to queries to the node APIs and storage back-ends. Should it not be located in a package related to that topic ?

src/aleph/sdk/models/post.py
```diff
@@ -0,0 +1,122 @@
+from datetime import datetime
+from typing import Any, Dict, Iterable, List, Optional, Union
+
+from aleph_message.models import Chain, ItemHash, ItemType, MessageConfirmation
+from pydantic import BaseModel, Field
+
+from .common import PaginationResponse, _date_field_to_float, serialize_list
+
+
+class Post(BaseModel):
```
hoh, line 10: Is this not present in `aleph-message` ?

src/aleph/sdk/models/post.py
```diff
@@ -0,0 +1,122 @@
...
+
+
+class PostsResponse(PaginationResponse):
+    """Response from an aleph.im node API on the path /api/v0/posts.json"""
+
+    posts: List[Post]
+    pagination_item = "posts"
+
+
+class PostFilter:
```
hoh, line 58: Same comment as for the other Filter class.

tests/integration/config.py
```diff
@@ -1,3 +1,3 @@
-TARGET_NODE = "http://163.172.70.92:4024"
+TARGET_NODE = "https://api1.aleph.im"
```
hoh, line 1: Why change this ? Is this related to this PR or should it be in another ?

src/aleph/sdk/base.py
MHHukiewitz: As @odesenfans pointed out, it is superfluous now, as aggregates are not being aggregated on each request (hence `limit` for how many aggregate messages into the past should be aggregated), but are materialized.

src/aleph/sdk/base.py
MHHukiewitz: Yeah, thought about this, probably the right call there

src/aleph/sdk/base.py
MHHukiewitz: It was already part of the few changes I had taken over. I thought, if it is something simple like this, I can just go all the way with all the strings.

src/aleph/sdk/models/message.py
MHHukiewitz: Check the structure on #63, based on #59, introduced a lot of functionality around the models, converting them to ORM (for the MessageCache) and back, so they all fit in nicely there.

src/aleph/sdk/models/common.py
MHHukiewitz: Agree on utility file, but regarding to docstrings, I think those are pretty self-explanatory and are extensively used in the `PostFilter` and `MessageFilter`

src/aleph/sdk/models/post.py
MHHukiewitz: Only `PostMessage` and `PostContent`

MHHukiewitz: But agree, after talking to @odesenfans I realized that there are a lot of data models that are in fact duplicated: Response models from the CCN API could be used in the SDK to parse them from the responses.

tests/integration/config.py
MHHukiewitz: Because the integration tests would fail and the IP is simply not reachable. The tests were never used, but I thought I could also fix them, while at it.

src/aleph/sdk/models/message.py
MHHukiewitz: My rationale here is: If I want to change anything about the "wiring" between requests/response types and the API, I'd find all the necessary code in the `aleph.sdk.models` package.

src/aleph/sdk/client.py
MHHukiewitz: IMHO, this is confusing

src/aleph/sdk/client.py
MHHukiewitz: Wondering if we should add `page` and `page_size` also as part of `MessageFilter` and `PostFilter`?

MHHukiewitz:
I refactored the package structure quite substantially here now, so not only we do now have a `model` module, containing all the Post, Message and Filter models, but also turned `client` into a module, following @hoh's suggestion.

hoh:
The diff looks good :+1: . Can you rebase the commits to have clear atomic changes that follow our [git commit style](https://pad.okeso.net/s/vWFVYWEuo#) ?

hoh (approved):

PR #71, opened 2023-10-25
Fix workflow triggers for Forked-Repo-PRs

MHHukiewitz:
Problem: build-wheels.yml and pytest-docker.yml would not trigger if a PR was created from a forked repo.

Solution: Trigger all the workflows on direct pushes to main and on any PR to main.


hoh:
We should run CI on all branches and Pull Requests.

What about using this ?
```
on:
  push:
  pull_request:
````

**Comment by MHHukiewitz** [comment]
> ```
> on:
>   push:
> [...]
> ```

Won't this cause the CI to trigger on every push on every branch, even without PR in place?

**Comment by hoh** [comment]
Good point.

The main thing that consumes our CI usage is the replication due to the matrix of versions.

We should run:
- `build-wheels.yml` on main and PRs that target main
- code quality on every push
- skip upgrading packages from the code quality on every push
- keep `pytest-docker.yml` on PR to main

File: .github/workflows/build-wheels.yml [hunk 1]
```diff
@@ -16,6 +22,14 @@ jobs:
         if: startsWith(matrix.os, 'ubuntu-')
         run: sudo echo RESET grub-efi/install_devices | sudo debconf-communicate grub-pc
 
+      - name: Cache dependencies
+        uses: actions/cache@v3
+        with:
+          path: ~/.cache/pip
+          key: ${{ runner.os }}-build-wheels-${{ hashFiles('setup.cfg', 'setup.py') }}
+          restore-keys: |
+            ${{ runner.os }}-build-wheels-
```
hoh, line 31: Why is this different from line 29 ?

MHHukiewitz: The `restore-keys` parameter is there to have a fallback on which to rely. So even if we do not have a cache hit, we can use the latest cached pip dependencies and just update them accordingly.

hoh (approved):

hoh (approved):

PR #73, opened 2023-10-25
Fix: Upload Test wasn't mocked

1yam:
## Problem
The workflow is failing sometime during this test due to `ServerDisconnectErrors`.

## Solutions
We can Mocked `AuthenticatedAlephHttpClient` to remove server call


MHHukiewitz (dismissed):
LGTM, but make sure that the test has the same syntax like before

MHHukiewitz (approved):
LGTM

PR #74, opened 2023-11-07
Fix: macos 11 & 12 build

1yam:


MHHukiewitz:
@hoh It seems that only with Python 3.11 the macOS wheel is able to successfully build. Is that fine?

MHHukiewitz (dismissed):
As we do not have a huge following on macOS and 3.11 is a reasonably recent Python release, the limitation should be ok. Should be mentioned though somewhere in the docs.

PR #75, opened 2023-11-09
Problem: Filtering by start_date or end_date would fail assertion

MHHukiewitz:
Solution: Cast to string in date field conversion


1yam (approved):
LGTM

1yam (approved):
LGTM

PR #87, opened 2023-12-07
Increase default VM memory to 2000 MiB and add validation

MHHukiewitz:
- Increased `DEFAULT_VM_MEMORY` to 2000
- Raise `ValueError` if validation of either `memory`, `vcpus` or `timeout` has invalid values.

Suggested value ranges:
- 2000 MiB <= `memory` <= 8000 MiB
- 1 <= `vcpus` <= 4
- 1 <= `timeout` <= 180


hoh:
As discussed in the meeting today, hardcoding such limits in the SDK does not make sense.

PR #90, opened 2024-01-17
* Custom Domain, raise exception at the end of checks

aliel:


hoh:
I got this to work and fix my issue, but the tests fail.

MHHukiewitz (approved):
Based on @hoh's comment and ChatGPTs explanation:

> **Old**: The method returned an error immediately when a mismatch was found.
> **New**: The method now evaluates all DNS rules and then checks if all values in status are False. If so, it raises an error. This implies that now the method completes checking all rules before deciding to raise an error.

Tests are passing, so LGTM

PR #92, opened 2024-01-30
[Fix] Loosen aleph-message dependency

MHHukiewitz:
Problem: Too strict aleph-message dependency

Solution: loosen it to accept compatible versions to 0.4.2

**EDIT:**
https://github.com/aleph-im/aleph-message/pull/91 fixed the backwards compatibility issue. `aleph-message` has been upgraded to 0.4.3 so these workflows pass again
> Another problem: Further problems arose from importing aleph-message~=0.4.2 due to some newer syntax that was not supported in Python 3.9 yet.
> 
> Another solution: EOL for Python 3.9 & out-of-the-box Ubuntu-20.04.
> 
> Also fixed deprecation warning for deprecated GH workflows that utilized EOL node.js v16 by using newer versions of those workflows.

### DIFF
--- a/.github/workflows/build-wheels.yml
+++ b/.github/workflows/build-wheels.yml
@@ -12,7 +12,7 @@ jobs:
   build:
     strategy:
       matrix:
-        os: [macos-11, macos-12, ubuntu-22.04, ubuntu-24.04]
+        os: [macos-11, macos-12, ubuntu-20.04, ubuntu-22.04]
     runs-on: ${{ matrix.os }}
 
     steps:
--- a/.github/workflows/build-wheels.yml
+++ b/.github/workflows/build-wheels.yml
@@ -19,7 +19,7 @@ jobs:
       - uses: actions/checkout@v4
 
       - name: Workaround github issue https://github.com/actions/runner-images/issues/7192
-        if: startsWith(matrix.os, 'ubuntu-')
+        if: startsWith(matrix.os, 'ubuntu-22')
         run: sudo echo RESET grub-efi/install_devices | sudo debconf-communicate grub-pc
 
       - name: Set up Python
--- a/.github/workflows/build-wheels.yml
+++ b/.github/workflows/build-wheels.yml
@@ -12,7 +12,7 @@ jobs:
   build:
     strategy:
       matrix:
-        os: [macos-11, macos-12, ubuntu-22.04]
+        os: [macos-11, macos-12, ubuntu-22.04, ubuntu-24.04]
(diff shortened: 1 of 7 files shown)


setup.cfg
```diff
@@ -38,7 +38,7 @@ install_requires =
     eciespy>=0.3.13; python_version>="3.11"
     typing_extensions
     typer
-    aleph-message==0.4.1
+    aleph-message~=0.4.2
```
nesitor, line 41:
```suggestion
    aleph-message~=0.4.1
```
This should be like that to accept `0.4.1` and `0.4.2` versions

MHHukiewitz:
Rebased on #93

hoh:
Commits mix different topics, can you ensure only the right commits are in this branch ?

nesitor (approved):

MHHukiewitz:
Test is failing as aleph-message=0.4.2 has syntax that is not supported anymore in Python 3.8. Should we drop support for that version?

hoh:
> Test is failing as aleph-message=0.4.2 has syntax that is not supported anymore in Python 3.8. Should we drop support for that version?

Yes, we can drop support for Python < 3.9

MHHukiewitz:
I don't know why the new check for ubuntu 24.04 isn't being picked up on... @hoh

.github/workflows/build-wheels.yml
```diff
@@ -12,24 +12,24 @@ jobs:
   build:
     strategy:
       matrix:
-        os: [macos-11, macos-12, ubuntu-20.04, ubuntu-22.04]
```
hoh, line 15: Can we keep support for Ubuntu 20.04 until it is EOL ? That is Apr 2025
See https://ubuntu.com/about/release-cycle

docker/ubuntu-24.04.dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ubuntu:20.04
+FROM ubuntu:24.04
```
hoh, line 1: Is that released already ?

.github/workflows/build-wheels.yml
MHHukiewitz: Yes, reverted that and fixed by upgrading to patched aleph-message~=0.4.3

docker/ubuntu-24.04.dockerfile
MHHukiewitz: Reverted

hoh (approved):

PR #93, opened 2024-01-30
[Lint] Reformat with latest `black` version

MHHukiewitz:
Problem: new version of black invalidates prior formatting

Solution: update black and reformat files

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -166,7 +166,7 @@ exclude =
     dist
     .eggs
     docs/conf.py
-ignore = E501 W291 W503 E203
+ignore = E501 W291 W503 E203 E704
 
 [isort]
 profile = black
--- a/examples/httpgateway.py
+++ b/examples/httpgateway.py
@@ -1,5 +1,6 @@
 """ Server metrics upload.
 """
+
 # -*- coding: utf-8 -*-
 
 import click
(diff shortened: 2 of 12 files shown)


nesitor (approved):

setup.cfg
```diff
@@ -166,7 +166,7 @@ exclude =
     dist
     .eggs
     docs/conf.py
-ignore = E501 W291 W503 E203
+ignore = E501 W291 W503 E203 E704
```
hoh, line 169: Why allow Multiple statements on one line (def) (E704) ?

https://www.flake8rules.com/rules/E704.html

MHHukiewitz: Because reformatting with the new version of black gives me this weird style:
```py
    def get_address(self) -> str: ...
```
Which flake8 doesn't approve of with E704
https://github.com/aleph-im/aleph-sdk-python/pull/93/files#diff-35dae2b4eb7a74cd319316827f4bcdbfd0e2267c489e4d5139433e6d250e6c99R21

MHHukiewitz: It might be a nice rule, but I haven't seen anyone doing weird stuff without this limitation, so I think it's fine to add it to ignore

hoh: Related to https://github.com/psf/black/issues/3887

PR #94, opened 2024-01-30
[Feature] Deprecation messages for invalid imports

MHHukiewitz:
Update Aleph SDK client to raise ImportError with appropriate error message for deprecated classes and types.

### DIFF
--- a/src/aleph/sdk/__init__.py
+++ b/src/aleph/sdk/__init__.py
@@ -16,12 +16,18 @@ __all__ = ["AlephHttpClient", "AuthenticatedAlephHttpClient"]
 
 def __getattr__(name):
     if name == "AlephClient":
-        raise ImportError("AlephClient has been turned into an abstract class. Please use `AlephHttpClient` instead.")
+        raise ImportError(
+            "AlephClient has been turned into an abstract class. Please use `AlephHttpClient` instead."
+        )
     elif name == "AuthenticatedAlephClient":
-        raise ImportError("AuthenticatedAlephClient has been turned into an abstract class. Please use `AuthenticatedAlephHttpClient` instead.")
+        raise ImportError(
+            "AuthenticatedAlephClient has been turned into an abstract class. Please use `AuthenticatedAlephHttpClient` instead."
+        )
     elif name == "synchronous":
-        raise ImportError("The 'aleph.sdk.synchronous' type is deprecated and has been removed from the aleph SDK. Please use `aleph.sdk.client.AlephHttpClient` instead.")
+        raise ImportError(
+            "The 'aleph.sdk.synchronous' type is deprecated and has been removed from the aleph SDK. Please use `aleph.sdk.client.AlephHttpClient` instead."
+        )
     elif name == "asynchronous":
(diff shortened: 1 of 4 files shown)


src/aleph/sdk/__init__.py
```diff
@@ -12,3 +12,14 @@
...
+
+def __getattr__(name):
+    if name == "AlephClient":
+        raise ImportError(f"AlephClient has been turned into an abstract class. Please use `AlephHttpClient` instead.")
+    elif name == "AuthenticatedAlephClient":
+        raise ImportError(f"AuthenticatedAlephClient has been turned into an abstract class. Please use `AuthenticatedAlephHttpClient` instead.")
+    elif name == "synchronous":
+        raise ImportError(f"The 'aleph.sdk.synchronous' type is deprecated and has been removed from the aleph SDK. Please use `aleph.sdk.client.AlephHttpClient` instead.")
+    elif name == "asynchronous":
+        raise ImportError(f"The 'aleph.sdk.asynchronous' type is deprecated and has been removed from the aleph SDK. Please use `aleph.sdk.client.AlephHttpClient` instead.")
```
hoh, line 25: Would it not be better to do a dictionary lookup ?

`ImportError` does not seem like the right exception here :thinking: 
[See https://docs.python.org/3/library/exceptions.html](https://docs.python.org/3/library/exceptions.html#exception-hierarchy)
`AttributeError`  is the standard class for this in Python, see 
https://docs.python.org/3/reference/datamodel.html#object.__getattr__

Why use f-strings without any formatting ?

```suggestion
        if name in DEPRECATION_MESSAGES:
            raise AttributeError(DEPRECATION_MESSAGES[name])
        else:
            return super().__getattr__(name)
```

nesitor (approved):

src/aleph/sdk/__init__.py
```diff
@@ -12,3 +12,22 @@
     del get_distribution, DistributionNotFound
 
 __all__ = ["AlephHttpClient", "AuthenticatedAlephHttpClient"]
+
+
+def __getattr__(name):
```
hoh, line 17: Can you explain why this code is there ?

MHHukiewitz: So that:

```py
def test_deprecation():
    with pytest.raises(ImportError):
        from aleph.sdk import AlephClient  # noqa

    with pytest.raises(ImportError):
        from aleph.sdk import AuthenticatedAlephClient  # noqa

    with pytest.raises(ImportError):
        from aleph.sdk import synchronous  # noqa

    with pytest.raises(ImportError):
        from aleph.sdk import asynchronous  # noqa

    with pytest.raises(ImportError):
        import aleph.sdk.synchronous  # noqa

    with pytest.raises(ImportError):
        import aleph.sdk.asynchronous  # noqa

    from aleph.sdk import AlephHttpClient  # noqa
```

hoh (approved):

hoh:
Needs squashing ;-)

MHHukiewitz:
A shit, I merged it with the GitHub Mobile App, and it just went the lazy route T.T

PR #98, opened 2024-02-06
Feature: Choose a ssl certificate other than the default one

Antonyjin:
I had the following problem:

```ssl:True [SSLCertVerificationError: (1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate```

when using the function: **AuthenticatedAlephHttpClient** and **AlephHttpClient**

I searched on the internet for a way to solve this problem, but all the commands/advice given didn't work.

So I thought it would be a good idea to give the user the option of specifying a specific SSL certificate if they wish.

This worked in my case and gave me the option of continuing to use the SDK provided by Aleph.

### DIFF
--- a/src/aleph/sdk/client/http.py
+++ b/src/aleph/sdk/client/http.py
@@ -1,8 +1,8 @@
 import json
 import logging
-import ssl
 from io import BytesIO
 from typing import Any, AsyncIterable, Dict, Iterable, List, Optional, Type
+import ssl
 
 import aiohttp
 from aleph_message import parse_message
@@ -31,12 +31,12 @@ class AlephHttpClient(AlephClient):
     http_session: aiohttp.ClientSession
 
     def __init__(
-            self,
-            api_server: Optional[str] = None,
-            api_unix_socket: Optional[str] = None,
-            allow_unix_sockets: bool = True,
-            timeout: Optional[aiohttp.ClientTimeout] = None,
-            ssl_context: Optional[ssl.SSLContext] = None,
+        self,
+        api_server: Optional[str] = None,
+        api_unix_socket: Optional[str] = None,
+        allow_unix_sockets: bool = True,
+        timeout: Optional[aiohttp.ClientTimeout] = None,
+        ssl_context: Optional[ssl.SSLContext] = None,
     ):
         """AlephClient can use HTTP(S) or HTTP over Unix sockets.
         Unix sockets are used when running inside a virtual machine,
@@ -83,7 +83,7 @@ class AlephHttpClient(AlephClient):
         params: Dict[str, Any] = {"keys": key}
 
         async with self.http_session.get(
-                f"/api/v0/aggregates/{address}.json", params=params
+            f"/api/v0/aggregates/{address}.json", params=params
         ) as resp:
             resp.raise_for_status()
             result = await resp.json()
@@ -91,7 +91,7 @@ class AlephHttpClient(AlephClient):
             return data.get(key)
 
     async def fetch_aggregates(
-            self, address: str, keys: Optional[Iterable[str]] = None
+        self, address: str, keys: Optional[Iterable[str]] = None
     ) -> Dict[str, Dict]:
         keys_str = ",".join(keys) if keys else ""
         params: Dict[str, Any] = {}
@@ -99,8 +99,8 @@ class AlephHttpClient(AlephClient):
             params["keys"] = keys_str
 
         async with self.http_session.get(
-                f"/api/v0/aggregates/{address}.json",
-                params=params,
+            f"/api/v0/aggregates/{address}.json",
+            params=params,
         ) as resp:
             resp.raise_for_status()
             result = await resp.json()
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by MHHukiewitz** [comment]
What was the context of your error? Did you develop an app that somehow has custom SSL handling? Could you share your code?

**Comment by hoh** [comment]
In addition to the comments from @MHHukiewitz , how can we ensure that this change keeps working ? 
Do you know how to add unit tests to test this feature automatically ?

**Comment by Antonyjin** [comment]
I tried to develop a very simple Python application with Flask. 
My goal was to try out some of the features of the Python SDK. I don't think this application uses a custom SSL.

Here's the code I ran: 

app.py
```py
import asyncio
from flask import Flask, render_template, request
from templates import script

app = Flask(__name__)


@app.route("/")
def index():
    return render_template("index.html")


@app.route("/submit", methods=["POST"])
def submit():
    if request.method == "POST":
        asyncio.run(script.create_agg())
        return render_template("index.html")
    return render_template("/")


@app.route("/get_aggregate", methods=["GET"])
def get_aggregate():
    if request.method == "GET":
        asyncio.run(script.get_agg())
        return render_template("index.html")
    return render_template("/")


@app.route("/post_file", methods=["POST"])
def post_file():
    file = request.files["file"]
    content = file.read()
    if request.method == "POST" and content != None:
        asyncio.run(script.post_file(content))
        return render_template("index.html")
    return render_template("/")

@app.route("/post_program", methods=["POST"])
def post_program():
    code = request.files["code"]
    if request.method == "POST" and code != None:
        asyncio.run(script.post_program(code))
        return render_template("index.html")
    return render_template("/")


if __name__ == "__main__":
    app.run(debug=True)
```

scripts.py
```py
from aleph_message.models import MessageType
from aleph.sdk.chains.ethereum import get_fallback_account
from aleph.sdk.client import AuthenticatedAlephHttpClient
from aleph.sdk.client import AlephHttpClient
from aleph.sdk.types import StorageEnum
from aleph.sdk.conf import settings as sdk_settings
from aleph_message.models.execution.program import Encoding
import ssl
import certifi

sslcontext = ssl.create_default_context(cafile=certifi.where())


async def create_agg():
        account = get_fallback_account()
        async with AuthenticatedAlephHttpClient(account, ssl_context=sslcontext) as client:
            message, status = await client.create_aggregate(
                "profile",
                {"bio": "tester", "name": "Antony"},
            )
            print(message.content)
            print(message.item_hash)
            print(status)


async def get_agg():
    async with AlephHttpClient(ssl_context=sslcontext) as client:
        aggregate = await client.fetch_aggregate(
            "0xbC80BeEcBd67549E70cb1C729e903818E6370D37",
            "profile",
        )
        print(aggregate)


async def post_file(file_content):
    account = get_fallback_account()
    async with AuthenticatedAlephHttpClient(account, ssl_context=sslcontext) as client:
        storage = StorageEnum.ipfs if len(file_content) > 4 * 1024 * 1024 else StorageEnum.storage
        message, status = await client.create_store(
            account.get_address(),
            file_content,
            storage_engine=storage,
        )
        print(message.item_hash)
        print(status)


async def post_program(file):
    account = get_fallback_account()

    async with AuthenticatedAlephHttpClient(account, ssl_context=sslcontext) as client:
        file_content = file.read()
        storage_engine = StorageEnum.ipfs if len(file_content) > 4 * 1024 * 1024 else StorageEnum.storage
        user_code, status = await client.create_store(
            file_content=file_content,
            storage_engine=storage_engine,
            guess_mime_type=True,
        )

        print(status)
        program_ref = user_code.item_hash
        message, status = await client.create_program(
            program_ref=program_ref,
            entrypoint="main:app",
            encoding=Encoding.squashfs,
            runtime=sdk_settings.DEFAULT_RUNTIME_ID,
            channel="default",
        )
        print(f"Program created successfully with item hash: {message.item_hash}")
        print(status)
```

Antonyjin:
I tested it on my app and it worked fine.
But I understand that I shouldn't rely on that alone. I'll find out so I can provide unit tests with it, if the change is relevant and accepted of course.

hoh:
The core issue is that apparently some systems running macOS have an issue providing CA Certificates to Python libraries. @Antonyjin reported that one of his friend could reproduce the problem while another could not.

Investigating the root cause of this issue may prove to be very time consuming, and @Antonyjin wanted to make something work within a reasonable timeframe. A solution was to use rely instead on the `certifi` library to provide CA Certificates authored by Mozilla.

In order for aiohttp to use a different set of CA, a custom version of the field `ssl_context` must be passed in argument.

src/aleph/sdk/client/http.py
```diff
@@ -48,6 +50,8 @@ def __init__(
         if unix_socket_path and allow_unix_sockets:
             check_unix_socket_valid(unix_socket_path)
             connector = aiohttp.UnixConnector(path=unix_socket_path)
+        elif ssl_context:
+            connector = aiohttp.TCPConnector(ssl=ssl_context)
```
MHHukiewitz, line 54: A potential issue I see here is that we create two connectors (first `UnixConnector` and then `TCPConnector`), if we pass `ssl_context` into the __init__ function, due to the way the if/elif is setup. You can avoid that with:
```suggestion
        if ssl_context:
            connector = aiohttp.TCPConnector(ssl=ssl_context)
        elif unix_socket_path and allow_unix_sockets:
            check_unix_socket_valid(unix_socket_path)
            connector = aiohttp.UnixConnector(path=unix_socket_path)
```

MHHukiewitz (changes requested):
LGTM, but if more parameters might be necessary to configure the connector in the future, it might be a good idea to think about just adding a parameter that just accepts any kind of `aiohttp.BaseConnector`, so that developers can configure them bythemselves: https://docs.aiohttp.org/en/stable/client_reference.html#connectors

MHHukiewitz:
Should look good, running CI/CD and if green, we merge :+1:

MHHukiewitz:
@Antonyjin best to run `flake8 src/ tests/ && mypy src / tests/ && isort src/ tests/ && black src/ tests/` in order to fix all the code quality problems

MHHukiewitz (dismissed):

Antonyjin:
Okay working on it!

MHHukiewitz (approved):
LGTM :+1:

PR #99, opened 2024-02-08
Fix: An ETH account could not be initialized from its mnemonic

hoh:
Users could not easily import or migrate accounts using their mnemonic representation.

Solution: Add a static method `from_mnemonic` on the `ETHAccount` class.

Discussion: This is a first step and this behaviour can be extended to more chains in the future.


MHHukiewitz (approved):
LGTM

PR #100, opened 2024-02-08
Fix: `create_instance` required a program `encoding`

hoh:
Problem: Instances don't use program encodings. This argument was left from refactoring.

Solution: Drop the argument.


MHHukiewitz (approved):
LGTM

PR #101, opened 2024-02-08
Fix: Payment could not be specified for instances

hoh:
Problem: Users of the SDK could not create instances with a specific payment method such as token streams.

Solution: Add a new argument, `payment`, to `AuthenticatedAlephClient` and use it in the "Instance" messages generated.

The argument is optional and defaults to "hold" on "ETH" for backward compatibility.

Discussion:
The argument is added just after mandatory arguments so it can be made mandatory in the future. This may however break backward compatibility with existing code that does not call the function using keywords arguments.


MHHukiewitz (approved):
LGTM

MHHukiewitz:
What ever the hell the AI is doing man :man_shrugging:

PR #108, opened 2024-03-13
Loosen eth-abi dependency

MHHukiewitz:
Problem: Too strict of a eth-abi==4.0.0b2 dependency makes it difficult to install the latest version of the SDK on aleph-nodestatus.
Screenshot from 2024-03-13 12-32-26


hoh (approved):

hoh (changes requested):
Why is the dependency present twice ?

If it is already in `install_requires`, no need to provide it in `[ethereum]`

PR #109, opened 2024-03-13
Remove double declaration of ethereum dependencies

MHHukiewitz:
Now only required if installed with `pip install aleph-sdk-python[ethereum]`.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -39,9 +39,6 @@ install_requires =
     typing_extensions
     typer
     aleph-message~=0.4.3
-    eth_account>=0.4.0
-    # Required to fix a dependency issue with parsimonious and Python3.11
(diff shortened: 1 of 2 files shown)


hoh:
Are we certain that neither of these libraries is used for something outside ethereum ?

MHHukiewitz:
Now I'm certain, `ETHAccount` was imported as a default value for `_load_account()` in `aleph.sdk.account`

MHHukiewitz:
There still seems to be some place where the ethereum dependencies are being imported by default in `aleph.sdk`, I will have to investigate, but I think it might have to do something with eciespy and coincurve being still present in the default installation. (See #110 for eciespy)

MHHukiewitz:
Closing this as important foundation messages are signed with ETH and might need to be validated in future.

Issue #37, opened 2023-07-10
Downloads should use files on disk instead of bytes in RAM


hoh:
Downloading files currently stores everything in RAM, doing so in a new `BytesIO` object.

We should change the definition of the function or create a new one that saves the data on disk.

My proposal is to ask the user to pass a file-like object the function can write to.

This file-like object can be a buffer in memory, a file on disk or anything compatible with the interface.

For convenience, we could make that argument optional and return a [NamedTemporaryFile](https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile) if it is None.

_Originally posted by @hoh in https://github.com/aleph-im/aleph-sdk-python/pull/36#discussion_r1258473666_

odesenfans:
What I told @1yam was to implement a generic function that takes a buffer, as you suggest. He already implemented it (see `download_file_to_buffer()`.

On the other hand, I don't see the point of returning a temporary file if no buffer is specified. I think we should provide a `download_to_file(file_hash, path)` convenience function that uses `download_file_to_buffer()`. I say we leave the choice of where the file is located entirely to the user.

PR #115, opened 2024-03-28
Add file_path param to `download_file()` to directly download files to disk

MHHukiewitz:
Instead of keeping them in memory. Addresses #37.

### DIFF
--- a/src/aleph/sdk/client/http.py
+++ b/src/aleph/sdk/client/http.py
@@ -1,7 +1,9 @@
 import json
 import logging
+import os.path
 import ssl
 from io import BytesIO
+from pathlib import Path
 from typing import Any, AsyncIterable, Dict, Iterable, List, Optional, Type, Union
 
 import aiohttp
@@ -209,17 +211,33 @@ class AlephHttpClient(AlephClient):
     async def download_file(
         self,
         file_hash: str,
-    ) -> bytes:
+        file_path: Optional[Union[Path, str]],
+    ) -> Optional[bytes]:
         """
         Get a file from the storage engine as raw bytes.
 
-        Warning: Downloading large files can be slow and memory intensive.
+        Warning: Downloading large files can be slow and memory intensive. Using the `file_path` parameter is encouraged to prevent storing the file in memory.
 
         :param file_hash: The hash of the file to retrieve.
+        :param file_path: The path to which the file should be saved.
+        :returns: The file's bytes, if no `file_path` is given, otherwise None.
         """
-        buffer = BytesIO()
-        await self.download_file_to_buffer(file_hash, output_buffer=buffer)
-        return buffer.getvalue()
+        if not file_path:
+            buffer = BytesIO()
+            await self.download_file_to_buffer(file_hash, output_buffer=buffer)
+            return buffer.getvalue()
+
+        if file_path is str:
+            file_path = Path(str)
+
+        if not os.path.exists(file_path):
+            dir_path = os.path.dirname(file_path)
+            if dir_path:
+                os.makedirs(dir_path, exist_ok=True)
+
+        with open(file_path, "wb") as file_buffer:
+            await self.download_file_to_buffer(file_hash, output_buffer=file_buffer)
+        return None
 
     async def download_file_ipfs(
         self,
--- a/tests/unit/test_download.py
+++ b/tests/unit/test_download.py
@@ -1,3 +1,7 @@
+import os.path
+import shutil
+from pathlib import Path
+
 import pytest
 
 from aleph.sdk import AlephHttpClient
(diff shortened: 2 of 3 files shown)


MHHukiewitz:
@hoh I think codecov is incorrectly configured. It currently also measures test file coverage... Which doesn't make sense.

hoh:
> @hoh I think codecov is incorrectly configured. It currently also measures test file coverage... Which doesn't make sense.

Why not ? Tests that don't run do not make sense.

MHHukiewitz:
> Why not ? Tests that don't run do not make sense.

Because I just added some cleanup checks after downloading to the file, in order to check the automatic directory and file creation every time correctly

tests/unit/test_download.py
```diff
@@ -15,10 +19,24 @@
...
+async def test_download_to_file():
+    download_path = "./downloads/test.txt"
+    if os.path.exists(download_path):
+        shutil.rmtree(Path(download_path).parent)
+    assert not os.path.exists(download_path)
+    async with AlephHttpClient(api_server=sdk_settings.API_HOST) as client:
+        await client.download_file(
+            "QmeomffUNfmQy76CQGy9NdmqEnnHU9soCexBnGU3ezPHVH", "./downloads/test.txt"
+        )
+    assert os.path.exists(download_path)
```
hoh, line 37: This test mutates the user's file system and implements a custom way to handle temporary files and directories.

Here are proposals for enhancements:

1. **Use `pathlib` for Path Management**: Replace `os.path.exists` and other path-related operations with `pathlib.Path` methods to make path handling more intuitive and Pythonic.
2. **Temporary Directory**: Utilize the `tempfile.TemporaryDirectory` context manager for automatic cleanup of the temporary directory. This ensures that the temporary directory and its contents are deleted automatically once the context is exited.

Here's the revised version of the code:

```python
async def test_download_to_file():
    with tempfile.TemporaryDirectory() as temp_dir:
        temp_dir_path = Path(temp_dir)
        download_path = temp_dir_path / "test.txt

        async with AlephHttpClient(api_server=sdk_settings.API_HOST) as client:
            await client.download_file(
                "QmeomffUNfmQy76CQGy9NdmqEnnHU9soCexBnGU3ezPHVH", str(download_path)
            )

        # Assert that the file now exists after download
        assert download_path.is_file()
```

src/aleph/sdk/client/http.py
```diff
@@ -209,17 +211,33 @@ async def download_file_ipfs_to_buffer(
...
-        buffer = BytesIO()
-        await self.download_file_to_buffer(file_hash, output_buffer=buffer)
-        return buffer.getvalue()
+        if not file_path:
+            buffer = BytesIO()
+            await self.download_file_to_buffer(file_hash, output_buffer=buffer)
+            return buffer.getvalue()
+
+        if file_path is str:
+            file_path = Path(file_path)
```
hoh, line 231: The code snippet provided demonstrates a common misunderstanding when attempting to check an object's type in Python:

```python
if file_path is str:
    file_path = Path(file_path)
```

Here are the reasons why this approach is incorrect:

1. **Incorrect Type Checking**: The expression `file_path is str` checks if `file_path` is the exact same object as the `str` class, which is not the intended use. This condition will always evaluate to `False` because `file_path` is an instance of a type (likely a string in this context), not the type (`str`) itself. The correct way to check if `file_path` is an instance of `str` is by using the `isinstance()` function, like so: `isinstance(file_path, str)`.

2. **Type Checking Pythonic Style**: In Python, it's often more idiomatic to embrace "duck typing" or to check for the presence of an attribute or behavior rather than explicitly checking an object's type. However, when you do need to check a type explicitly, `isinstance()` is the preferred way.

3. **Usage of `Path` with Strings**: The intention behind the conditional block seems to be converting a string file path into a `Path` object for more convenient path manipulation using the `pathlib` module.

Why not just write `file_path = Path(file_path)` ? If it's already a Path, it should be fine as well.

src/aleph/sdk/client/http.py
```diff
@@ -209,17 +211,33 @@ async def download_file_ipfs_to_buffer(
     async def download_file(
         self,
         file_hash: str,
-    ) -> bytes:
+        file_path: Optional[Union[Path, str]] = None,
+    ) -> Optional[bytes]:
```
hoh, line 215: Is the same function either returning bytes or saving data in a file and returning None ?
That looks like a very weird interface... Would it not be better to have an additional method lik `async def download_file_to(path) -> pathlib.Path` instead ?

hoh (changes requested):
The approach for dual use of the `download_file` seems very confusing, see my comment on the topic.

This branch is re-implementing in it a lot of patterns that are already covered by the Python standard library.

src/aleph/sdk/client/http.py
MHHukiewitz: Yeah, that was my first idea. I also don't like it that we can't have polymorphism-like selection of signatures, which would've been ideal in this case.

tests/unit/test_download.py
```diff
@@ -19,6 +22,23 @@ async def test_download(file_hash: str, expected_size: int):
         assert file_size == expected_size
 
 
+@pytest.mark.asyncio
+async def test_download_to_file():
```
1yam, line 26: Maybe we could Mock the AlephHTTPClient as in [Upload_test](https://github.com/aleph-im/aleph-sdk-python/blob/1c2d57d847a5865bc92c93a4cfaf74d122e2b20d/tests/unit/test_upload.py#L14)

Don't thinks we really need to get a real file on the network for this test

1yam (approved):
LGTM

PR #120, opened 2024-04-23
Fix: Could not generate signed messages outside HTTP client

hoh:
Problem: Developers could not generate signed messages without using the AlephHttpClient.

Solution: Move and rename the method on the abstract class; Move the utility that generates sha256 hashes to utils.py.


MHHukiewitz (approved):
LGTM

PR #123, opened 2024-06-10
Fix: 'called_once' is not a valid assertion.

hoh:
AttributeError: 'called_once' is not a valid assertion. Use a spec for the mock if 'called_once' is meant to be an attribute.


1yam (approved):
LGTM

PR #129, opened 2024-06-14
Fix: aiohttp issues a DeprecationWarning

hoh:
For passing bytes to a file upload instead of a BytesIO


src/aleph/sdk/client/authenticated_http.py
```diff
@@ -135,7 +136,7 @@ async def storage_push_file(self, file_content) -> str:
         Push a file to the storage service.
         """
         data = aiohttp.FormData()
-        data.add_field("file", file_content)
+        data.add_field("file", BytesIO(file_content))
```
Psycojoker, line 139: You can add the `bytes` type declaration in the function signature for `file_content` here.

Also, I know it's outside of this PR but the return type of this function doesn't seems 100% correct: we can get a `None` from the `dict.get`.

I'm surprised that mypy didn't catched this error, is it correctly configured? https://github.com/aleph-im/aleph-sdk-python/actions/runs/9562286305/job/26358345937?pr=129#step:7:15 

I remember something about not having to specify `None` from now on but it was for arguments iirc but I might be mistaking.

Same comment for the return type of the function `ipfs_push_file` above.

Psycojoker (dismissed):
Except the small possible improvements, LGTM.

PR #133, opened 2024-06-21
This branch adds a dockerized environment to reproduce the issues that the CI faced with Ubuntu 24.04 and secp256k1.

hoh:
Usage:
```sh
bash ./tests/test-ubuntu-24.04-docker.sh
```

Then use the upper arrow to run tests with `hatch`.


hoh:
See https://github.com/baking-bad/pytezos/issues/370

hoh:
Upstream to fix this in https://github.com/baking-bad/pytezos/pull/373/files

PR #137, opened 2024-07-02
Add new Confidential fields

nesitor:
Problem: As a user we cannot create a confidential VM using the SDK.

Solution: Implement new confidential VM fields.

### DIFF
--- a/src/aleph/sdk/client/abstract.py
+++ b/src/aleph/sdk/client/abstract.py
@@ -27,7 +27,7 @@ from aleph_message.models import (
     PostMessage,
     parse_message,
 )
-from aleph_message.models.execution.environment import HypervisorType
+from aleph_message.models.execution.environment import HostRequirements, HypervisorType
 from aleph_message.models.execution.program import Encoding
 from aleph_message.status import MessageStatus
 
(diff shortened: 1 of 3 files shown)


src/aleph/sdk/client/abstract.py
```diff
@@ -395,10 +395,13 @@ async def create_instance(
         internet: bool = True,
         aleph_api: bool = True,
         hypervisor: Optional[HypervisorType] = None,
+        confidential_firmware: Optional[ItemHash] = None,
```
hoh, line 398: Why not call this simply `firmware` ?
```suggestion
        firmware: Optional[ItemHash] = None,
```

src/aleph/sdk/client/abstract.py
```diff
@@ -395,10 +395,13 @@ async def create_instance(
...
+        confidential_policy: Optional[int] = None,
```
hoh, line 399: I would assume that this policy field is specific to AMD SEV. 
Would it make sense to name it `sev_policy` instead ?

src/aleph/sdk/client/abstract.py
nesitor: Because it's the firmware specific for the confidential instances.

src/aleph/sdk/client/abstract.py
nesitor: This policy field isn't specific for AMD, as on the `aleph-message` schema it isn't, I just followed the same schema specifications.

hoh: Here is how the competitor, Intel TXD, is configured: https://github.com/intel/tdx-linux/wiki/Instruction-to-set-up-TDX-host-and-guest.

There is no notion of `policy` there, the policy number is specific to AMD.

Psycojoker (approved):
I can't tell for the business logic part really well yet, but at least from the python part this PR looks good to me :+1:

olethanh (approved):

PR #143, opened 2024-08-05
Feature: get_program_price

1yam:
Problem:
In the Pay-As-You-Go (PAYG) flow, we need to retrieve the price of a program.

Solution:
Implement the get_program_price function to query the `/api/v0/price/` endpoint and fetch the price of the program.

### DIFF
--- a/src/aleph/sdk/client/http.py
+++ b/src/aleph/sdk/client/http.py
@@ -409,5 +409,5 @@ class AlephHttpClient(AlephClient):
                 )
             except aiohttp.ClientResponseError as e:
                 if e.status == 400:
-                    raise InvalidHashError(f"No such hash {item_hash}")
+                    raise InvalidHashError(f"Bad request or no such hash {item_hash}")
                 raise e
--- a/src/aleph/sdk/client/http.py
+++ b/src/aleph/sdk/client/http.py
@@ -12,9 +12,14 @@ from aleph_message.models import AlephMessage, ItemHash, ItemType
 from pydantic import ValidationError
 
 from ..conf import settings
-from ..exceptions import FileTooLarge, ForgottenMessageError, MessageNotFoundError
+from ..exceptions import (
+    FileTooLarge,
+    ForgottenMessageError,
+    MessageNotFoundError,
+    InvalidHashError,
+)
 from ..query.filters import MessageFilter, PostFilter
-from ..query.responses import MessagesResponse, Post, PostsResponse
+from ..query.responses import MessagesResponse, Post, PostsResponse, PriceResponse
 from ..types import GenericMessage
 from ..utils import (
     Writable,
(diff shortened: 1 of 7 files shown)


src/aleph/sdk/client/http.py
```diff
@@ -392,3 +397,17 @@ async def watch_messages(
                         yield parse_message(data)
                 elif msg.type == aiohttp.WSMsgType.ERROR:
                     break
+
+    async def get_program_price(self, item_hash: str) -> PriceResponse:
```
hoh, line 401: What guarantees do we have that the `item_hash` corresponds to a program ?

src/aleph/sdk/client/http.py
```diff
@@ -392,3 +397,17 @@ async def watch_messages(
...
+        async with self.http_session.get(f"/api/v0/price/{item_hash}") as resp:
+            try:
+                resp.raise_for_status()
+                response_json = await resp.json()
+                return PriceResponse(
+                    required_tokens=response_json["required_tokens"],
+                    payment_type=response_json["payment_type"],
+                )
+            except aiohttp.ClientResponseError as e:
+                if e.status == 400:
```
hoh, line 411: 400 == Bad Request, should this not be 404 ?

tests/unit/test_price.py
```diff
@@ -0,0 +1,40 @@
...
+from aleph.sdk.query.responses import PriceResponse
+
+
+def make_mock_price_client(item_hash: str):
+    mock_client = AsyncMock(spec=AlephHttpClient)
+    if item_hash == "valid_item_hash":
+        mock_client.get_program_price.return_value = PriceResponse(
+            required_tokens=3.0555555555555556e-06, payment_type="superfluid"
+        )
+    elif item_hash == "invalid_item_hash":
```
hoh, line 16: Why not two different mocks ?

tests/unit/test_price.py
```diff
@@ -0,0 +1,40 @@
...
+        mock_client.get_program_price.side_effect = InvalidHashError("Invalid hash")
+    else:
+        raise NotImplementedError(f"Mock not implemented for item_hash: {item_hash}")
+    return mock_client
+
+
+@pytest.mark.asyncio
+async def test_get_program_price_valid():
```
hoh, line 24: What is this test testing ? Can you add a docstring ? Would it not be better to mock the response of the HTTP API ?

src/aleph/sdk/client/http.py
1yam: I made it based on what api is returning me, and it give me a 400.
```
aiohttp.client_exceptions.ClientResponseError: 400, message='Bad Request', url=URL('https://api2.aleph.im/api/v0/price/b')
```

src/aleph/sdk/client/http.py
1yam: We don't really have one, but in all case it's gonna check server side,
So is it making sense to check if it an executable message one time  on client side and one time on server side ?

olethanh (approved):

PR #146, opened 2024-08-14
fix: aiohttp.ClientSession needs to be created inside an async function

Psycojoker:
This should fix the CI, it used to be a deprecated warning before but this is now an error.

Idk if this changes the way you are supposed to use this class and if it could break the existing API :/


olethanh (approved):
not sure this is the best way but I have tested a few command with the http client and it still work properly

PR #150, opened 2024-08-16
Fix: confest _http_session instead of http_session

1yam:

PR #151, opened 2024-08-19
fix: use str instead of float to preserve precision 

1yam:
This PR addresses an issue where using float to represent certain numerical values resulted in loss of precision due to rounding. By switching to str, we can maintain the exact value without rounding errors.


1yam:
Unit test should work need [PR](https://github.com/aleph-im/aleph-sdk-python/pull/150) merge

hoh:
Converted to Draft, this should use the Decimal type in Python and ensuring that the upstream API returns the expected type.

src/aleph/sdk/query/responses.py
```diff
@@ -77,5 +77,5 @@ class MessagesResponse(PaginationResponse):
 class PriceResponse(BaseModel):
     """Response from an aleph.im node API on the path /api/v0/price/{item_hash}"""
 
-    required_tokens: float
+    required_tokens: str
```
hoh, line 80: `pyaleph` should expose the information as a string, and every client should handle it given its own properties (ex: Python = Decimal)

```suggestion
    required_tokens: Decimal
```

hoh: image

hoh: We agreed in the meeting of today to use the integer version as a string, a field containing the number of decimals of the token, and the (low-precision) float representation of the value.

Psycojoker:
Apparently the other PR has been merged, should we resume this PR?

PR #157, opened 2024-08-27
Fix warnings, remove BASE until superfluid upgrade

philogicae:
- Fix annoying warnings by defining eth-utils==4.1.1
- Disable BASE chain until superfluid.py repo is updated

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -33,6 +33,7 @@ dependencies = [
     "aioresponses>=0.7.6",
     "superfluid~=0.2.1",
     "eth_typing==4.3.1",
+    "eth-utils==4.1.1",
     "web3==6.3.0",
 ]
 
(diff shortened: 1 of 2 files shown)


Psycojoker:
What is the warning?

philogicae:
> What is the warning?

```bash
/home/philogicae/.pyenv/versions/3.11.8/lib/python3.11/site-packages/eth_utils/network.py:44: UserWarning: Network 345 with name 'Yooldo Verse Mainnet' does not have a valid ChainId. eth-typing should be updated with the latest networks.
  warnings.warn(
/home/philogicae/.pyenv/versions/3.11.8/lib/python3.11/site-packages/eth_utils/network.py:44: UserWarning: Network 12611 with name 'Astar zkEVM' does not have a valid ChainId. eth-typing should be updated with the latest networks.
  warnings.warn(
 ```
 It appears since we added web3py to the SDK (was needed).
 It's very ugly when using the CLI, since it shows up at the beginning of every command.
 Now I have a dep conflict with ledgereth, which is an old package to support Ledger keys, currently trying to find a better/newer solution.

philogicae:
Fine after all

PR #158, opened 2024-08-27
Fix optional trusted_exec

philogicae:
Problem: when creating non-coco instances, trusted_execution field is populated
Solution: To not include it

Note: CCN will still put "trusted_execution": null in the final processed message.


olethanh (changes requested):
This doesn't change anything,  if you pass trusted_execution=None or just don't pass it to InstanceEnvironment

You can test it by yourself by creating an InsatanceContent and using the method .dict(exclude_none=True)

olethanh:
Normally that bug with the trusted_execution was already fixed and the client already didn't send the trusted_execution field

olethanh:
In that client commit https://github.com/aleph-im/aleph-client/commit/9eba48d5f95d41ca819db467649a78bfddb26b02

philogicae:
Canceled

PR #160, opened 2024-08-28
Fix _load_account when both string & path private key are provided

philogicae:
Check title


olethanh (changes requested):
That's not a bug, that a warning because both shouldn't be provided at the same time.

olethanh (approved):
After discussion with @philogicae  I was wrong since the SDK provide a file by default it was failing when we passed a string  adress on the client.

PR #162, opened 2024-08-29
Fix settings

philogicae:


1yam (dismissed):
LGTM

src/aleph/sdk/conf.py
```diff
@@ -78,6 +114,16 @@ class Settings(BaseSettings):
...
+    # Add all placeholders to allow easy dynamic setup of CHAINS
+    locals().update(
+        {
+            f"CHAINS_{chain.upper() if isinstance(chain, str) else Chain(chain).value}_{key.upper()}": (
+                value if value is not None else ""
+            )
+            for chain, info in CHAINS.items()
+            for key, value in info
+        }
+    )
```
hoh, line 126: Are we sure we want to use `locals()` here ? It's a hidden feature of the language for meta-programming, I am not a big fan of using it here... :thinking:

hoh: This would also likely not work with Mypy and other validation tools.

hoh (approved):

PR #163, opened 2024-08-30
Feature: get_message_status and overload of get_message to return status

1yam:
Problem: Currently, there is no method to retrieve the status of a message using its item_hash. This limitation makes it difficult to efficiently check the processing state of specific messages.

Solution: I have added a new method, get_message_status, which allows users to obtain the status of a message by providing the item_hash. I have updated the get_message method to optionally return the status of a message when requested, improving overall flexibility and functionality.

### DIFF
--- a/src/aleph/sdk/client/http.py
+++ b/src/aleph/sdk/client/http.py
@@ -18,6 +18,7 @@ from typing import (
 )
 
 import aiohttp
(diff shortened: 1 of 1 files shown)


philogicae (approved):
LGTM

hoh (commented):
No test on this ?

PR #165, opened 2024-09-14
Fix: dependencies were pinned too strictly

hoh:
This prevented installing the SDK on some systems
such as Nix 24.05.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -32,8 +32,8 @@ dependencies = [
     "typing_extensions",
     "aioresponses>=0.7.6",
     "aleph-superfluid>=0.2.1",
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -32,8 +32,8 @@ dependencies = [
     "typing_extensions",
     "aioresponses>=0.7.6",
     "aleph-superfluid>=0.2.1",
-    "eth_typing==4.3.1",
-    "web3==6.3.0",
```
Psycojoker, line 36: We want to add an upper bound version to enforce semver to avoid randomly breaking our installation.

Psycojoker (changes requested):

PR #175, opened 2024-09-30
Feature: Account Handler

1yam:
Description:

This PR introduces a new configuration file chains_config.json and updates the _load_account function to handle Solana accounts. These changes simplify the process of importing wallets for Solana users.

Changes:

Added: Support for Solana accounts in the _load_account function.
Created: chains_config.json for chain-specific configurations.
chains_config.json Format:
Added: Helper functions to streamline wallet importation and account management

json
```
{
   "path": "path",
   "chain": "chain"
}
```

### DIFF
--- a/src/aleph/sdk/account.py
+++ b/src/aleph/sdk/account.py
@@ -131,8 +131,8 @@ def _load_account(
     elif private_key_path and private_key_path.is_file():
         if private_key_path:
             try:
-                # Look for the account by private_key_path in CHAINS_CONFIG_FILE
-                with open(settings.CHAINS_CONFIG_FILE, "r") as file:
+                # Look for the account by private_key_path in CONFIG_FILE
+                with open(settings.CONFIG_FILE, "r") as file:
                     accounts = json.load(file)
 
                 matching_account = next(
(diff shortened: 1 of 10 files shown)


src/aleph/sdk/conf.py
```diff
@@ -162,6 +168,19 @@ class Config:
...
+    # If Config file exist and well filled we update the PRIVATE_KEY_FILE default
+    if settings.CONFIG_FILE.exists():
+        try:
+            with open(settings.CONFIG_FILE, "r", encoding="utf-8") as f:
+                config_data = json.load(f)
+
+            if "path" in config_data:
+                settings.PRIVATE_KEY_FILE = Path(config_data["path"])
+        except json.JSONDecodeError:
+            pass
```
nesitor, line 182: These code fragment should be outside the code, because we will load the config also if the user changed to use a path different from the default one.

src/aleph/sdk/account.py
```diff
@@ -33,10 +123,55 @@ def _load_account(
...
+            except KeyError as e:
+                logger.error(f"Missing key in account config: {e}")
+                raise ValueError(
+                    f"Invalid account data in {private_key_path}. Key {e} is missing."
+                )
+            except Exception as e:
+                logger.error(f"Error loading account from {private_key_path}: {e}")
+                raise ValueError(
+                    f"Could not load account data from {private_key_path}."
+                )
```
nesitor, line 173: This logic isn't needed as the user specifies which `account_type` and which `private_key` path he wants, so we only need to load it.
If you want to provide auto-detect options, you need to put the `account_type` argument as Optional and load the current chain that the user selected.

tests/unit/test_utils.py
```diff
@@ -235,3 +239,107 @@ def test_compute_confidential_measure():
         )
         == b"ls2jv10V3HVShVI/RHCo/a43WO0soLZf0huU9ZZstIw="
     )
+
+
+def test_parse_solana_private_key_bytes():
```
nesitor, line 244: These test should be moved to solana test, not utils test file.

nesitor (changes requested):
There are a lot of things to change and improve, let's do a pair session to fix all these things.

src/aleph/sdk/conf.py
nesitor: We will do in a future improvement.

nesitor (approved):

PR #176, opened 2024-10-02
feat(linting): add pydantic to allow mypy to do better testing

Psycojoker:
I've discovered while adding yamlfix that does pull pydantic on #172 that
adding it actually allows mypy to test more things.

Turns out those "more things" are actually type errors in our code.

This WIP PR is here to fix those.

PR #179, opened 2024-10-15
177 upgrade pydantic version

Antonyjin:


pyproject.toml
```diff
@@ -30,15 +30,16 @@ dynamic = [ "version" ]
...
   "aleph-superfluid>=0.2.1",
-  "base58==2.1.1",                         # Needed now as default with _load_account changement
+  "base58==2.1.1",                                                          # Needed now as default with _load_account changement
   "coincurve; python_version<'3.11'",
   "coincurve>=19; python_version>='3.11'",
   "eth-abi>=4; python_version>='3.11'",
   "eth-typing==4.3.1",
   "jwcrypto==1.5.6",
-  "pynacl==1.5",                           # Needed now as default with _load_account changement
+  "pydantic-settings>=2",
```
hoh, line 41: What is this dependency ?

src/aleph/sdk/chains/common.py
```diff
@@ -72,7 +72,7 @@ async def sign_message(self, message: Dict) -> Dict:
         """
         message = self._setup_sender(message)
         signature = await self.sign_raw(get_verification_buffer(message))
-        message["signature"] = signature.hex()
+        message["signature"] = "0x" + signature.hex()
```
hoh, line 75: Why is this added here ?

src/aleph/sdk/chains/remote.py
```diff
@@ -52,7 +52,7 @@ async def from_crypto_host(
             session = aiohttp.ClientSession(connector=connector)
 
         async with session.get(f"{host}/properties") as response:
-            response.raise_for_status()
+            await response.raise_for_status()
```
hoh, line 55: Is this related to pydantic ? Should this use `await` or should the implementation not be async ?

src/aleph/sdk/chains/remote.py
```diff
@@ -75,7 +75,7 @@ def private_key(self):
     async def sign_message(self, message: Dict) -> Dict:
         """Sign a message inplace."""
         async with self._session.post(f"{self._host}/sign", json=message) as response:
-            response.raise_for_status()
+            await response.raise_for_status()
```
hoh, line 78: Same: Is this related to pydantic ? Should this use await or should the implementation not be async ?

src/aleph/sdk/client/http.py
```diff
@@ -467,3 +467,6 @@ async def get_message_status(self, item_hash: str) -> MessageStatus:
             if resp.status == HTTPNotFound.status_code:
                 raise MessageNotFoundError(f"No such hash {item_hash}")
             resp.raise_for_status()
+
+        data = await resp.json()
+        return MessageStatus(**data)
```
hoh, line 472: Was there no `return` before ?

tests/unit/aleph_vm_authentication.py
```diff
@@ -63,36 +63,33 @@ class SignedPubKeyHeader(BaseModel):
...
+    @field_validator("payload")
     def payload_must_be_hex(cls, value: bytes) -> bytes:
         """Convert the payload from hexadecimal to bytes"""
-
         return bytes_from_hex(value.decode())
 
-    @root_validator(pre=False, skip_on_failure=True)
-    def check_expiry(cls, values) -> Dict[str, bytes]:
+    @model_validator(mode="after")
+    def check_expiry(cls, values: "SignedPubKeyHeader") -> "SignedPubKeyHeader":
```
hoh, line 77: Why quotes on the type annotations ? What about `from __futures__ import annotations` ?

pyproject.toml
Antonyjin: image

By moving to pydantic v2, the settings management has been moved to a separate package called pydantic-settings so if we don't import it, it causes some errors

src/aleph/sdk/chains/remote.py
Antonyjin: This is not related to pydantic, just saw some warning when lauching the tests, so I tried to fix some of them because it wasn't a lot of work.

src/aleph/sdk/chains/remote.py
Antonyjin: Same: This is not related to pydantic, just saw some warning when lauching the tests, so I tried to fix some of them because it wasn't a lot of work.

tests/unit/aleph_vm_authentication.py
Antonyjin: Completely forgot about this. I've added it to the code, it works fine and it will be included in the next push.

src/aleph/sdk/client/http.py
Antonyjin: You're right. When I initially tested upgrading to Pydantic v2, I encountered an error saying that a return value was required, so I added it. I just checked again, and the error is no longer there, so I may have done it incorrectly. I'll remove it in the next push.

src/aleph/sdk/chains/common.py
Antonyjin: Apologies, my mistake. Normally, `.hex()` should add the `0x` at the beginning of the signature, but after the changes made for the Pydantic upgrade, it no longer does. It’s just missing this, so I added it temporarily to check if everything would pass and completely forgot to correct it.

I'll debug to see why the `hex()` don't add the `0x` at the begening

Antonyjin: I found out that with python3>=3.5 .hex() should remove the 0x prefix from the string.
My branch does it so it causes some errors in the tests but on the main branch it seems that the prefix is not removed even though the python version is 3.11.5

philogicae:
Added [`Add pydantic for better mypy tests + Fixes`](https://github.com/aleph-im/aleph-sdk-python/pull/179/commits/3e88f28cdebf0f2c3a173653ab0a803174693ab2)
Related to [ALEPH-196](https://aleph-im.atlassian.net/browse/ALEPH-196) and this [PR](https://github.com/aleph-im/aleph-sdk-python/pull/176)

philogicae (approved):
LGTM.
Note: Check needed on the "signature" unresolved conversation @Antonyjin

nesitor (dismissed):

nesitor (approved):

PR #182, opened 2024-10-31
Implement new EVM chains

nesitor:


src/aleph/sdk/chains/evm.py
```diff
@@ -0,0 +1,48 @@
...
+from .common import get_fallback_private_key
+from .ethereum import ETHAccount
+
+
+class EVMAccount(ETHAccount):
+    def __init__(self, private_key: bytes, chain: Optional[Chain] = None):
+        super().__init__(private_key, chain)
+        # Decide if we have to send also the specified chain value or always use ETH
+        # if chain:
+        #     self.CHAIN = chain
```
1yam, line 17: We need this check, else on the message we will only have `ETH`.

nesitor: Yes, but for now it's fine to continue sending the messages with ETH chain as they are EVMs and the signature doesn't change.

philogicae (approved):

PR #184, opened 2024-11-06
Fix wrong account config alert

nesitor:
Wrong account config alert where shown loading the default account:
wrong alert

This PR fixes that wrong warning message.


philogicae (approved):
LGTM

PR #191, opened 2024-12-06
Implement GPU support

nesitor:
Problem: If a user wants to deploy a VM using the GPU support, he cannot do it from the SDK because it's not updated with latest aleph_message protocol.

Solution: Update aleph_message protocol with last changes.


olethanh (dismissed):
need to be upgraded once aleph-message is released ofc but otherwhise lgtm

philogicae (dismissed):

philogicae (approved):

PR #192, opened 2024-12-09
fix(create_instance): Default memory in docstring

RezaRahemtola:
Fixed to match the value of `settings.DEFAULT_INSTANCE_MEMORY` that is used in the code.


Psycojoker (approved):
Thx!

PR #203, opened 2025-02-27
Fix: Remove references to slow server api1.aleph.im

hoh:
Server api1.aleph.im is very slow and outdated
(Core i7 from 2018, up to 40 seconds to respond
to `/metrics` in the monitoring).

We suspect that this causes issues in the
monitoring and performance of the network.

This branch removes all references to api1 and
replaces them with api3 where relevant.


1yam (approved):
LGTM

PR #225, opened 2025-08-04
Feature: Voucher integrations

1yam:
Related to ALEPH-542

This pull request introduces a new voucher service for managing and retrieving vouchers across EVM and Solana chains, along with tests for the functionality. The most significant changes include the creation of `AuthenticatedVoucher` and `Vouchers` classes, the addition of voucher-related types, configuration updates, and comprehensive unit tests.

### Voucher Service Implementation:

- **`src/aleph/sdk/client/services/authenticated_voucher.py`:** Added the `AuthenticatedVoucher` class, which extends `Vouchers`. It allows retrieval of vouchers without explicitly passing an address by using the account address as a fallback. This includes methods for fetching vouchers (`get_vouchers`, `get_evm_vouchers`, `get_solana_vouchers`) and resolving addresses.
- **`src/aleph/sdk/client/services/voucher.py`:** Introduced the `Vouchers` class for fetching voucher data from EVM and Solana chains. It includes utilities for resolving addresses, fetching metadata, and retrieving vouchers by chain.

### Voucher Types and Configuration:

- **`src/aleph/sdk/types.py`:** Added `Voucher`, `VoucherMetadata`, and `VoucherAttribute` types to represent voucher data and metadata. These types encapsulate attributes like name, description, external URL, image, and traits.
- **`src/aleph/sdk/conf.py`:** Added configuration settings for voucher metadata URLs, Solana registry URLs, and the voucher sender address. These settings are used by the voucher services to interact with external resources.

### Unit Tests:

- **`tests/unit/services/test_authenticated_voucher.py`:** Added tests for the `AuthenticatedVoucher` class, verifying address resolution logic and voucher retrieval for both EVM and Solana chains. Tests ensure fallback mechanisms work correctly when no address is provided.
- **`tests/unit/services/test_voucher.py`:** Added tests for the `Vouchers` class, covering metadata fetching, voucher retrieval by chain, and chain detection logic for both EVM and Solana addresses.

### DIFF
--- a/src/aleph/sdk/client/http.py
+++ b/src/aleph/sdk/client/http.py
@@ -568,7 +568,7 @@ class AlephHttpClient(AlephClient):
             message, status = await self.get_message(
                 item_hash=ItemHash(item_hash), with_status=True
             )
-            if status != MessageStatus.PROCESSED:
+            if status not in [MessageStatus.PROCESSED, MessageStatus.REMOVING]:
                 resp = f"Invalid message status: {status}"
             elif message.type != MessageType.store:
                 resp = f"Invalid message type: {message.type}"


src/aleph/sdk/client/http.py
```diff
@@ -61,6 +61,7 @@
     safe_getattr,
 )
 from .abstract import AlephClient
+from .services.voucher import Vouchers
```
nesitor, line 64: I think this import should be in the same form than other ones below:
https://github.com/aleph-im/aleph-sdk-python/pull/225/files#diff-98b2f220317fe0379eef848b853313affd30595845e3c4483da9fdb54cdc3a4bR36-R40

src/aleph/sdk/conf.py
```diff
@@ -92,6 +92,12 @@ class Settings(BaseSettings):
     )
     SCHEDULER_URL: ClassVar[str] = "https://scheduler.api.aleph.cloud/"
 
+    VOUCHER_METDATA_TEMPLATE_URL: str = (
+        "https://claim.twentysix.cloud/sbt/metadata/{}.json"
+    )
+    VOUCHER_SOL_REGISTRY: str = "https://api.claim.twentysix.cloud/v1/registry/sol"
+    VOUCHER_SENDER: str = "0xB34f25f2c935bCA437C061547eA12851d719dEFb"
```
nesitor, line 99: I think that instead `VOUCHER_SENDER` should be something like `VOUCHER_ORIGIN_ADDRESS`
```suggestion
    VOUCHER_ORIGIN_ADDRESS: str = "0xB34f25f2c935bCA437C061547eA12851d719dEFb"
```

tests/unit/services/test_authenticated_voucher.py
nesitor, line 1: These test are useless because are mocking everything. Instead to be useful needs to instantiate a real Account and check that method returns the result expected.

tests/unit/services/test_voucher.py
```diff
@@ -0,0 +1,112 @@
...
+    MOCK_METADATA,
+    MOCK_SOLANA_ADDRESS,
+    MOCK_SOLANA_REGISTRY,
+    MOCK_VOUCHER_ID,
+)
+
+
+@pytest.mark.asyncio
+async def test_get_evm_vouchers(mock_post_response, make_mock_aiohttp_session):
+    mock_client = MagicMock()
```
nesitor, line 19: Why mocking the entire client? Instead just mock the method you need to get the results to ensure that there is not an error on the instantiation or other calls.

tests/unit/services/test_voucher.py
```diff
@@ -0,0 +1,112 @@
...
+async def test_get_solana_vouchers(make_mock_aiohttp_session):
+    mock_client = MagicMock()
+    voucher_service = Vouchers(client=mock_client)
+
+    registry_session = make_mock_aiohttp_session(MOCK_SOLANA_REGISTRY)
+    metadata_session = make_mock_aiohttp_session(MOCK_METADATA)
+
+    with patch(
+        "aiohttp.ClientSession", side_effect=[registry_session, metadata_session]
+    ):
```
nesitor, line 43: Same as above, if you already mock the aiohttp call, is not needed to mock the Client.

nesitor (commented):
The are some things to change on tests and also one or 2 needed on the code.

tests/unit/services/test_authenticated_voucher.py
1yam: Well logic of Voucher is tested,

I mock the account to match the `.get_address` to match the the address, with what `get_posts` will give.

The logic of voucher is still tested since:
- We mocked the get_posts to return mocked_data
- We mocked the accounts (where it's only to get address.get_address) to match the mocked address
- We mocked the http request to metdata / solana voucher

But we do not mocked any logic of voucher they still process just with mocked data

tests/unit/services/test_voucher.py
nesitor: It's not a good pattern to mock everything, we need to mock only elements not involved on the logic we want to test, like http responses, calls to database, etc.

src/aleph/sdk/client/http.py
```diff
@@ -565,7 +568,7 @@ async def get_stored_content(
             message, status = await self.get_message(
                 item_hash=ItemHash(item_hash), with_status=True
             )
-            if status not in [MessageStatus.PROCESSED, MessageStatus.REMOVING]:
+            if status != MessageStatus.PROCESSED:
```
nesitor, line 571: I guess that we should maintain both status handling.

PR #237, opened 2025-09-30
Fix:  `credit_balances` does not exist

1yam:
Fix issue:
1) credit_balance got replaced by credit_history & unit test mocked replaced with more cohérent value (from staging instead of fake data generated by claude)


nesitor (approved):
🚀

PR #238, opened 2025-10-02
Solve Ledger Derivation path signing

nesitor:
Fix: Solved wrong derivation path on message signing from Ledger device.


1yam (approved):
LGTM

PR #279, opened 2026-02-21
Add new methods to manage instance server (reinstall, backup create, info, download, delete )

aliel:
### This PR add new methods to perform new actions on an instance

- reinstall_instance to reinstall an existing instance to its initial state
- create_backup to create a backup of an instance and return a pre-signed download link
- get_backup to get backup info if it exists and return a pre-signed download link
- restore_from_file to restore an instance from a provided rootfs file
- restore_from_volume to restore an instance from a pinned volume

related to:
https://github.com/aleph-im/aleph-vm/pull/874

### DIFF
--- a/src/aleph/sdk/client/vm_client.py
+++ b/src/aleph/sdk/client/vm_client.py
@@ -1,6 +1,7 @@
 import datetime
 import json
 import logging
+from pathlib import Path
 from typing import Any, AsyncGenerator, Dict, List, Optional, Tuple
 from urllib.parse import urlparse
 
@@ -106,7 +107,11 @@ class VmClient:
         return domain
 
     async def perform_operation(
-        self, vm_id: ItemHash, operation: str, method: str = "POST"
+        self,
+        vm_id: ItemHash,
+        operation: str,
+        method: str = "POST",
+        params: Optional[Dict[str, str]] = None,
     ) -> Tuple[Optional[int], str]:
         if not self.pubkey_signature_header:
             self.pubkey_signature_header = (
@@ -119,7 +124,7 @@ class VmClient:
 
         try:
             async with self.session.request(
-                method=method, url=url, headers=header
+                method=method, url=url, headers=header, params=params
             ) as response:
                 response_text = await response.text()
                 return response.status, response_text
@@ -193,6 +198,109 @@ class VmClient:
 
     async def erase_instance(self, vm_id: ItemHash) -> Tuple[Optional[int], str]:
         return await self.perform_operation(vm_id, "erase")
+
+    async def reinstall_instance(
+        self, vm_id: ItemHash, erase_volumes: bool = True
+    ) -> Tuple[Optional[int], str]:
+        params = None if erase_volumes else {"erase_volumes": "false"}
+        return await self.perform_operation(vm_id, "reinstall", params=params)
+
+    async def create_backup(
+        self,
+        vm_id: ItemHash,
+        include_volumes: bool = False,
+        skip_fsfreeze: bool = False,
+        run_async: bool = False,
+    ) -> Tuple[Optional[int], str]:
+        params: Dict[str, str] = {}
+        if run_async:
+            params["async"] = "true"
+        if include_volumes:
+            params["include_volumes"] = "true"
+        if skip_fsfreeze:
+            params["skip_fsfreeze"] = "true"
+        return await self.perform_operation(vm_id, "backup", params=params or None)
+
+    async def get_backup(self, vm_id: ItemHash) -> Tuple[Optional[int], str]:
+        return await self.perform_operation(vm_id, "backup", method="GET")
+
+    async def delete_backup(
+        self, vm_id: ItemHash, backup_id: str
+    ) -> Tuple[Optional[int], str]:
+        if not self.pubkey_signature_header:
+            self.pubkey_signature_header = (
+                await self._generate_pubkey_signature_header()
+            )
+        url, header = await self._generate_header(
+            vm_id=vm_id, operation=f"backup/{backup_id}", method="DELETE"
+        )
+        try:
+            async with self.session.delete(url, headers=header) as response:
+                text = await response.text()
+                return response.status, text
+        except aiohttp.ClientError as e:
+            logger.error(f"HTTP error during backup delete: {str(e)}")
+            return None, str(e)
+
+    async def get_restore_endpoint(self, vm_id: ItemHash) -> Tuple[str, Dict[str, str]]:
+        """Return authenticated (url, headers) for a restore POST.
+
+        Use this when you need control over the upload (e.g. progress
+        tracking). For a simple restore, use restore_from_file instead.
+        """
+        if not self.pubkey_signature_header:
+            self.pubkey_signature_header = (
+                await self._generate_pubkey_signature_header()
+            )
+        return await self._generate_header(
+            vm_id=vm_id, operation="restore", method="POST"
+        )
+
+    async def restore_from_file(
+        self, vm_id: ItemHash, rootfs_path: str
+    ) -> Tuple[Optional[int], str]:
+        if not self.pubkey_signature_header:
+            self.pubkey_signature_header = (
+                await self._generate_pubkey_signature_header()
+            )
+        url, header = await self._generate_header(
+            vm_id=vm_id, operation="restore", method="POST"
+        )
+        data = aiohttp.FormData()
+        data.add_field(
+            "rootfs",
+            open(rootfs_path, "rb"),
+            filename=Path(rootfs_path).name,
+            content_type="application/octet-stream",
+        )
+        try:
+            async with self.session.post(url, headers=header, data=data) as response:
+                text = await response.text()
+                return response.status, text
+        except aiohttp.ClientError as e:
+            logger.error(f"HTTP error during restore: {str(e)}")
+            return None, str(e)
+
+    async def restore_from_volume(
+        self, vm_id: ItemHash, volume_ref: str
+    ) -> Tuple[Optional[int], str]:
+        if not self.pubkey_signature_header:
+            self.pubkey_signature_header = (
+                await self._generate_pubkey_signature_header()
+            )
+        url, header = await self._generate_header(
+            vm_id=vm_id, operation="restore", method="POST"
+        )
+        header["Content-Type"] = "application/json"
+        try:
+            async with self.session.post(
+                url, headers=header, json={"volume_ref": volume_ref}
+            ) as response:
+                text = await response.text()
+                return response.status, text
+        except aiohttp.ClientError as e:
+            logger.error(f"HTTP error during restore: {str(e)}")
+            return None, str(e)
 
     async def expire_instance(self, vm_id: ItemHash) -> Tuple[Optional[int], str]:
         return await self.perform_operation(vm_id, "expire")
--- a/src/aleph/sdk/client/vm_client.py
+++ b/src/aleph/sdk/client/vm_client.py
@@ -124,7 +124,9 @@ class VmClient:
 
         try:
             request_kwargs: Dict[str, Any] = {
-                "method": method, "url": url, "headers": header,
+                "method": method,
+                "url": url,
+                "headers": header,
             }
             if params:
                 request_kwargs["params"] = params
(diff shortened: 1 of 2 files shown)


odesenfans (changes requested):

foxpatch-aleph (commented):
The PR is well-structured and correct. The VmOperation enum is a clean improvement, the auth-header deduplication is a good simplification, backup_id validation prevents path traversal, and test coverage is solid for the happy paths. A few minor issues: restore_from_file opens the file with synchronous I/O in an async function (problematic for large rootfs images), some tests use regex URL matchers but don't assert that the expected query params were actually sent, and the get_backup name is ambiguous given it returns info rather than file content.

**`src/aleph/sdk/client/vm_client.py`** (line 278): Using synchronous `open()` (and aiohttp reading the file object synchronously) will block the event loop for large rootfs images. Consider either wrapping the upload in `asyncio.get_event_loop().run_in_executor(None, ...)` or using `aiofiles` to open and stream the file asynchronously:

```python
import aiofiles

async with aiofiles.open(rootfs_path, "rb") as f:
    data = aiohttp.FormData()
    data.add_field("rootfs", f, filename=rootfs_path.name, content_type="application/octet-stream")
    async with self.session.post(url, headers=header, data=data) as response:
        ...
```

For rootfs images (which can easily be several GB) this matters.

**`src/aleph/sdk/client/vm_client.py`** (line 291): `restore_from_file` catches `aiohttp.ClientError` but not `OSError` (e.g. `FileNotFoundError`). An `OSError` from the `open()` call will propagate to the caller without any logging. Either document that file-not-found raises, or catch and log it consistently with the `ClientError` branch.

**`src/aleph/sdk/client/vm_client.py`** (line 241): `get_backup` is ambiguous — based on the PR description it returns backup *info* (metadata + pre-signed download link), not a backup file. A name like `get_backup_info` would be more explicit and prevent confusion with `restore_from_file`.

**`tests/unit/test_vm_client.py`** (line 334): `test_reinstall_instance_keep_volumes` uses a regex matcher to accept any URL under `.../reinstall` but doesn't assert that `erase_volumes=false` was actually included in the query string. The test would pass even if the param was silently dropped. Consider capturing the real request and checking `request.url.query` (or using `aioresponses` `requests` dict to inspect the call arguments).

**`tests/unit/test_vm_client.py`** (line 378): Same issue as `test_reinstall_instance_keep_volumes`: `test_create_backup_with_options` doesn't verify that `include_volumes=true` and `skip_fsfreeze=true` were actually sent as query params. The regex match only confirms the right endpoint was called.

**`tests/unit/test_vm_client.py`** (line 498): Missing negative-path test for `restore_from_file`: no test covers what happens when `rootfs_path` does not exist. Since `OSError` is not caught (see comment above), this would raise an unhandled exception — worth documenting and testing explicitly.

foxpatch-aleph (commented):
The PR is well-structured: the `VmOperation` enum cleanly replaces raw strings, the `backup_id` path-traversal guard is correct, and the refactor to centralise auth-header generation in `_generate_header` removes duplication. Three quality concerns are worth addressing before merge: blocking synchronous file I/O inside an async function, an inconsistent error-handling surface in `restore_from_file`, and tests that don't actually verify the query parameters they exist to cover.

**`src/aleph/sdk/client/vm_client.py`** (line 278): **Blocking I/O in async context.** `open()` is synchronous and will stall the event loop for the entire read time. For a rootfs image (which can be several GB) this will freeze all other async tasks. Consider wrapping the upload in `asyncio.get_event_loop().run_in_executor(None, ...)` or using `aiofiles` for the file handle. Alternatively, stream the file with `aiohttp.streamer` so only small chunks are read at a time.

**`src/aleph/sdk/client/vm_client.py`** (line 291): **Inconsistent error surface.** `aiohttp.ClientError` is caught and returned as `(None, str(e))`, but `OSError` / `FileNotFoundError` (file doesn't exist, permission denied, etc.) propagates as a raw exception. Callers currently have to handle two completely different error shapes depending on the failure mode. Either catch `OSError` in the same block and return `(None, str(e))`, or let both bubble as exceptions and remove the catch entirely — but pick one convention.

**`src/aleph/sdk/client/vm_client.py`** (line 221): **Implicit reliance on server default for `erase_volumes=True`.** When `erase_volumes=True` (the default), no query parameter is sent at all; the server's own default is assumed to be `true`. If the server default ever changes this SDK method silently changes behaviour. Being explicit — always sending `erase_volumes=true` or `erase_volumes=false` — would be more robust:
```python
params = {"erase_volumes": "true" if erase_volumes else "false"}
```

**`tests/unit/test_vm_client.py`** (line 335): **Test doesn't verify the query parameter is actually sent.** `re.compile(rf"http://localhost/control/machine/{vm_id}/reinstall")` matches any URL with that prefix, including one with no query string at all. The test would pass even if the `erase_volumes=false` param were dropped. Assert against the recorded request instead:
```python
req = list(m.requests.values())[0][0]
assert req.kwargs["params"] == {"erase_volumes": "false"}
```
The same applies to `test_create_backup_with_options` — it uses a regex match but never checks that `include_volumes=true&skip_fsfreeze=true` appear in the request.

**`src/aleph/sdk/client/vm_client.py`** (line 241): **Naming nit: `get_backup` is ambiguous.** The PR description says it returns a pre-signed download link, but the name reads like it fetches the backup payload itself. `get_backup_info` or `backup_info` would be clearer about what is returned.

**`src/aleph/sdk/client/vm_client.py`** (line 318): **`manage_instance` still accepts `List[str]`** while all other call-sites now use `VmOperation`. This is pre-existing code touched by the refactor, so not a blocker, but it is the one remaining place that bypasses the enum and could accept an invalid operation string silently.

foxpatch-aleph (changes requested):
The new backup/restore/reinstall methods are well-structured, the VmOperation enum is a nice cleanup, and the path-traversal guard on backup_id is correct. However there is one genuine bug in restore_from_file (OSError from open() is not caught, so file-not-found or permission failures escape as uncaught exceptions rather than the (None, str) return the signature promises), and several test gaps that leave real behaviour unverified.

**`src/aleph/sdk/client/vm_client.py`** (line 278): Bug: `open(rootfs_path, "rb")` is inside the `try` block but only `aiohttp.ClientError` is caught. A missing file or permission error raises `OSError` (e.g. `FileNotFoundError`), which propagates uncaught to the caller. This is inconsistent with every other error-path in this class, which returns `(None, str(e))`. Either broaden the except clause (`except (aiohttp.ClientError, OSError)`), or let it raise and document that the caller must handle it — but the current state silently violates the return-type contract.

```python
except (aiohttp.ClientError, OSError) as e:
    logger.error(f"Error during restore: {e}")
    return None, str(e)
```

**`src/aleph/sdk/client/vm_client.py`** (line 221): Asymmetric param passing: when `erase_volumes=True` (the default) no query param is sent at all, relying on the server's default. When `False`, `erase_volumes=false` is sent explicitly. This is only correct if the server's default is to erase volumes. It's fragile — a server-side default change would silently break the SDK default. Consider always sending the param:

```python
params = {"erase_volumes": str(erase_volumes).lower()}
```

**`src/aleph/sdk/client/vm_client.py`** (line 232): Minor: `params` is initialised as `{}` and conditionally populated. An empty dict is falsy, so `if params:` in `perform_operation` will correctly skip it — but passing `{}` instead of `None` is confusing to readers who might not immediately recall that falsy check. Consider `params: Optional[Dict[str, str]] = None` and building it only when non-empty, or just pass `params or None`.

**`src/aleph/sdk/client/vm_client.py`** (line 39): The `__str__` override is redundant. `VmOperation` already inherits from `str`, so `str(VmOperation.STOP)` already returns `"stop"` without the override. Safe to remove.

**`src/aleph/sdk/client/vm_client.py`** (line 278): Synchronous `open()` inside an async function will block the event loop while the OS opens the file descriptor (normally negligible, but worth noting for large-file or slow-filesystem scenarios). Using `aiofiles` or wrapping in `asyncio.get_event_loop().run_in_executor` would be the idiomatic fix, though for an upload helper this is low priority.

**`tests/unit/test_vm_client.py`** (line 334): The test verifies the HTTP status but never asserts that `erase_volumes=false` was actually present in the outgoing request. The `aioresponses` mock matches on the URL prefix regex, so the test would pass even if the param were silently dropped. Add an assertion on the recorded request, e.g.:

```python
request_key = ("POST", URL(f"http://localhost/control/machine/{vm_id}/reinstall"))
# or inspect m.requests to check the query string
assert any("erase_volumes=false" in str(k[1]) for k in m.requests)
```

**`tests/unit/test_vm_client.py`** (line 498): Missing test: `restore_from_file` with a path that does not exist. Once the `OSError` handling is fixed, add a test that calls the method with a non-existent path and asserts `status is None` and the error string is returned (rather than an uncaught exception).

foxpatch-aleph (approved):
This PR adds comprehensive new methods for VM instance management (reinstall, backup/restore operations). The implementation is clean, follows existing patterns, and includes proper security validation (backup ID regex to prevent path traversal). The `VmOperation` enum is a nice improvement for type safety. Tests are thorough, covering all new methods and edge cases including invalid inputs and error handling.

**`src/aleph/sdk/client/vm_client.py`** (line 233): Minor optimization: the `params` variable is initialized to `None` but then only set if options are provided. Could simplify to: `params = {} if include_volumes or skip_fsfreeze else None` and then conditionally add keys.

**`src/aleph/sdk/client/vm_client.py`** (line 242): The method name `get_backup` is a bit ambiguous - it returns backup info via GET request. Consider renaming to `get_backup_info` or `list_backups` for clarity, since `create_backup` already exists for creating backups.

**`tests/unit/test_vm_client.py`** (line 386): The test `test_create_backup_with_options` doesn't verify that the query parameters are actually sent. Consider adding an assertion like `assert any("include_volumes=true" in str(k[1]) for k in m.requests)` similar to the reinstall tests.

**`tests/unit/test_vm_client.py`** (line 493): The test `test_restore_from_volume` doesn't verify that the `volume_ref` is sent in the JSON body. Consider adding a check that the request body contains the correct volume_ref.

foxpatch-aleph (changes requested):
The PR adds useful new functionality for instance management (reinstall, backup, restore operations) with good test coverage. The VmOperation enum is a nice addition for type safety, and the backup ID validation is good for security. However, there are a few issues that need to be addressed: (1) restore_from_file uses synchronous file I/O which blocks the event loop for large files, (2) manage_instance method should use the new VmOperation enum for consistency, and (3) the test_create_backup_with_options test doesn't verify that query parameters are actually sent.

**`src/aleph/sdk/client/vm_client.py`** (line 275): The `restore_from_file` method uses synchronous `with open()` in an async function. For large rootfs files, this will block the event loop. Consider using `aiofiles` for async file reading, or at least add a note about this limitation. Example:

```python
import aiofiles

async with aiofiles.open(rootfs_path, "rb") as f:
    # ... read in chunks or pass the file handle
```

Alternatively, if the file is small enough, this may be acceptable, but it should be documented.

**`src/aleph/sdk/client/vm_client.py`** (line 315): The `manage_instance` method still takes `List[str]` for operations instead of using the new `VmOperation` enum. For consistency with the rest of the codebase (which now uses `VmOperation`), this should be updated to accept `List[Union[VmOperation, str]]` or just `List[VmOperation]`. This would provide type safety and prevent typos in operation names.

**`tests/unit/test_vm_client.py`** (line 370): The `test_create_backup_with_options` test doesn't verify that the query parameters (`include_volumes=true&skip_fsfreeze=true`) are actually sent in the request. Similar to `test_reinstall_instance`, add an assertion like:

```python
assert any("include_volumes=true" in str(k[1]) for k in m.requests)
assert any("skip_fsfreeze=true" in str(k[1]) for k in m.requests)
```

**`src/aleph/sdk/client/vm_client.py`** (line 157): Minor: The `operate` method is a lower-level API that still uses string operations. Consider either updating it to use `VmOperation` or adding a docstring note that it's a lower-level method for advanced use cases.

foxpatch-aleph (changes requested):
The PR adds useful VM management features with good test coverage and security validation for backup_id. However, the volume_ref parameter in restore_from_volume lacks validation, posing a potential security risk. Additionally, restore_from_file uses blocking file I/O in an async context which could impact performance for large files.

**`src/aleph/sdk/client/vm_client.py`** (line 292): The `volume_ref` parameter lacks validation. Consider adding a regex validation similar to `_BACKUP_ID_RE` to prevent potential injection attacks. The backend may use this value directly in API calls.

**`src/aleph/sdk/client/vm_client.py`** (line 274): Using synchronous `open()` in an async function blocks the event loop. For large rootfs files, this could cause performance issues. Consider using `aiofiles.open()` for async file I/O, or at minimum add a docstring warning about this limitation.

**`tests/unit/test_vm_client.py`** (line 478): Add a test case for `restore_from_volume` with potentially malicious `volume_ref` values (e.g., containing path traversal or special characters) to ensure proper validation is in place.

PR #286, opened 2026-04-20
feat: add reserve_resources method for CRN capacity pre-check

aliel:


foxpatch-aleph (changes requested):
The PR adds a useful `reserve_resources` method for CRN capacity pre-check and refactors `create_vm_control_payload` to use a new `create_control_payload` helper. However, there are two issues: (1) the original `create_vm_control_payload` function had a bug where it created a `payload` dict but returned a different dict entirely - this was silently broken before and the PR correctly fixes it; (2) there's no test coverage for the new functionality. The code quality is good and follows existing patterns.

**`src/aleph/sdk/utils.py`** (line 240): The original `create_vm_control_payload` function had a subtle bug: it created a `payload` dict but then returned a completely new dict literal (the `return payload` at the end was unreachable). The PR correctly fixes this by extracting the common logic into `create_control_payload`. This is a good refactor that fixes a latent bug.

**`src/aleph/sdk/client/vm_client.py`** (line 337): The new `reserve_resources` method is missing test coverage. Please add tests following the existing pattern in `test_vm_client.py` (e.g., `test_perform_operation`, `test_notify_allocation`). Test both success (200) and failure (503) cases.

**`src/aleph/sdk/client/vm_client.py`** (line 368): Minor: The logger format string uses `%s` which works but could be more explicit. Consider using `str(e)` for consistency with the return statement, or use `.format()` or f-string: `logger.error("HTTP error during reserve_resources: %s", str(e))`

**`src/aleph/sdk/client/vm_client.py`** (line 327): Note: The `notify_allocation` method doesn't have error handling for `aiohttp.ClientError` while `reserve_resources` does. Consider adding similar error handling to `notify_allocation` for consistency, or document why it's different.

foxpatch-aleph (approved):
This PR adds a well-implemented `reserve_resources` method for CRN capacity pre-checks. The implementation follows existing authentication patterns, the refactoring of `create_vm_control_payload` to use a generic helper reduces duplication, and tests cover the main success and failure scenarios. The code is clean and consistent with the codebase.

**`src/aleph/sdk/client/vm_client.py`** (line 368): Minor: The `%s` format specifier will call `str()` automatically, so `str(e)` is redundant. Consider simplifying to `logger.error("HTTP error during reserve_resources: %s", e)` for consistency with typical logging patterns.

**`tests/unit/test_vm_client.py`** (line 270): Consider adding a test for the `aiohttp.ClientError` exception path that returns `(None, str(e))`. While this mirrors the pattern in other methods like `perform_operation`, adding coverage would improve test completeness.

PR #288, opened 2026-05-18
fix(crn): skip malformed CRN entries in CrnList.from_api

aliel:
A malformed CRN payload (e.g. missing GPU.model) previously raised
  ValidationError and crashed any consumer iterating the list, including
  `aleph instance list`. A hostile CRN operator could exploit this to
  break clients across the network.

  Validate each entry independently, log a warning identifying the
  offender, and skip non-dict entries. Add tests covering malformed,
  non-dict, and empty payloads.

### DIFF
--- a/src/aleph/sdk/client/services/crn.py
+++ b/src/aleph/sdk/client/services/crn.py
@@ -1,3 +1,4 @@
+import logging
 from datetime import datetime
 from typing import TYPE_CHECKING, Dict, List, Optional, Union
 
@@ -5,7 +6,7 @@ import aiohttp
 from aiohttp.client_exceptions import ClientResponseError
 from aleph_message.models import ItemHash
 from packaging.version import InvalidVersion, Version
-from pydantic import BaseModel, NonNegativeInt, PositiveInt
+from pydantic import BaseModel, NonNegativeInt, PositiveInt, ValidationError
 
 from aleph.sdk.conf import settings
 from aleph.sdk.exceptions import MethodNotAvailableOnCRN, VmNotFoundOnHost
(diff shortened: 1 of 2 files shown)


foxpatch-aleph (dismissed):
This PR hardens CrnList.from_api against malformed CRN payloads by wrapping model_validate in a try/except, logging a warning with the offending entry's identifier, and continuing. The fix is correct, minimal, and well-tested. No security issues, no logic errors. Ready to merge.

**`src/aleph/sdk/client/services/crn.py`** (line 136): Nit: `item.get("hash")` treats an empty string hash as falsy and falls through to `name`. Consider `item.get("hash") is not None` if an empty string hash should be treated as a valid identifier. No practical impact — just a consistency observation.

foxpatch-aleph (dismissed):
Correctly fixes a real availability issue where a single malformed CRN entry (e.g., missing GPU.model) would crash the entire CrnList.from_api() consumer. The try/except approach with per-entry validation, warning logging, and identifier extraction is appropriate. Tests cover malformed dict entries, non-dict items, None entries, and empty payloads. The code is clean and the fix is minimal — no over-engineering.

**`src/aleph/sdk/client/services/crn.py`** (line 11): Minor style nit: logger definition sits between imports (between line 9 and line 13). The existing code already has this pattern, so it's consistent with the codebase, but worth noting.

foxpatch-aleph (approved):
A clean, well-targeted fix for a real availability concern: malformed CRN entries returned by the CRN list endpoint no longer crash consumers like `aleph instance list`. Each entry is validated independently in a loop, with ValidationErrors caught and logged with a best-effort identifier. The test suite covers malformed entries, non-dict items, None values, and empty payloads. No correctness, security, or code quality issues.

odesenfans (approved):

